PR #22, opened 2020-04-01
Config consolidation

dhaavi:


dhaavi:
Merging early because this is blocking other development.

PR #24, opened 2020-04-06
Add support for PORTMASTER_DATA environment variable

ppacher:
This PR adds support for the PORTMASTER_DATA environment variable. It can be used instead of specifying `--data` for each invocation of pmctl/portmaster-control. This PR also updates the README to use `--data` instead of the deprecated `--db`.

### DIFF
--- a/pmctl/main.go
+++ b/pmctl/main.go
@@ -61,12 +61,12 @@ func init() {
 	// Let cobra ignore if we are running as "GUI" or not
 	cobra.MousetrapHelpText = ""
 
-	rootCmd.PersistentFlags().StringVar(&dataDir, "data", "", "set data directory")
-	rootCmd.PersistentFlags().StringVar(&databaseDir, "db", "", "alias to --data (deprecated)")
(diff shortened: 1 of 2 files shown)


pmctl/main.go
```diff
@@ -61,12 +61,12 @@ func init() {
 	// Let cobra ignore if we are running as "GUI" or not
 	cobra.MousetrapHelpText = ""
 
-	rootCmd.PersistentFlags().StringVar(&dataDir, "data", "", "set data directory")
-	rootCmd.PersistentFlags().StringVar(&databaseDir, "db", "", "alias to --data (deprecated)")
+	rootCmd.PersistentFlags().StringVar(&dataDir, "data", "", "Configures the data directory. It's also possible to configure the PORTMASTER_DATA environment variable.")
```
dhaavi, line 64: Description somewhat unclear, would recommend:
`Configures the data directory. Alternatively, this can also be set via the environment variable PORTMASTER_DATA.

PR #27, opened 2020-04-07
Rework Communication+Link to Connection

dhaavi:


firewall/master.go
```diff
@@ -31,91 +31,170 @@ import (
...
 	}
 
-	// grant self
-	if comm.Process().Pid == os.Getpid() {
-		log.Infof("firewall: granting own communication %s", comm)
-		comm.Accept("")
-		return
+	// check if process is communicating with itself
+	if pkt != nil {
+		if conn.Process().Pid >= 0 && pkt.Info().Src.Equal(pkt.Info().Dst) {
```
ppacher, line 51: What's that `conn.Process().Pid >= 0` check for? Btw, what about process A listening on `127.0.0.100` (localhost) and communicating with itself using a new socket. Src would be 127.0.0.1 and Dst 127.0.0.100?

firewall/master.go
```diff
@@ -31,91 +31,170 @@ import (
...
+				otherProcess, err := process.GetOrFindPrimaryProcess(pkt.Ctx(), otherPid)
+				if err == nil {
+
+					if otherProcess.Pid == conn.Process().Pid {
+						conn.Accept("connection to self")
+						return
+					}
+				}
+			}
+		}
```
ppacher, line 72: `process.GetPidByEndpoints` and `process.GetOrFindPrimaryProcess` can fail for various reasons and we completely hide those errors here. I guess there are only a few (if not only one) that we can safely hide. May add some logging for the others.

firewall/master.go
```diff
@@ -31,91 +31,170 @@ import (
...
-	// FIXME: comm.Entity.Unlock()
+	// check conn type
+	switch conn.Scope {
+	case network.IncomingHost, network.IncomingLAN, network.IncomingInternet, network.IncomingInvalid:
+		if p.BlockInbound() {
+			if conn.Scope == network.IncomingHost {
+				conn.Block("inbound connections blocked")
+			} else {
+				conn.Deny("inbound connections blocked")
+			}
```
ppacher, line 90: May add  a comment why we need to `Block` for `IncomingHost` and not `Deny`

firewall/master.go
```diff
@@ -31,91 +31,170 @@ import (
...
+		default: // netutils.Invalid
+			conn.Deny("invalid IP") // Block Outbound / Drop Inbound
+			return
+		}
+	} else if conn.Entity.Domain != "" {
+		// DNS Query
+		// DNS is expected to resolve to LAN or Internet addresses
+		// TODO: handle domains mapped to localhost
+		if p.BlockScopeInternet() && p.BlockScopeLAN() {
```
ppacher, line 129: Do we only block the connection if both, `ScopeInternet` **and** `ScopeLAN are marked for blocking?

firewall/master.go
```diff
@@ -31,91 +31,170 @@ import (
...
 }
 
 // FilterDNSResponse filters a dns response according to the application profile and settings.
-func FilterDNSResponse(comm *network.Communication, q *resolver.Query, rrCache *resolver.RRCache) *resolver.RRCache { //nolint:gocognit // TODO
+func FilterDNSResponse(conn *network.Connection, q *resolver.Query, rrCache *resolver.RRCache) *resolver.RRCache { //nolint:gocognit // TODO
 	// do not modify own queries - this should not happen anyway
-	if comm.Process().Pid == os.Getpid() {
+	if conn.Process().Pid == os.Getpid() {
 		return rrCache
 	}
```
ppacher, line 190: If this is supposed to not happen we should log it!

network/connection.go
```diff
@@ -0,0 +1,379 @@
...
+		// check if it already exists
+		mapKey := strconv.Itoa(conn.process.Pid) + "/" + conn.Scope
+		dnsConnsLock.RLock()
+		_, ok := dnsConns[mapKey]
+		dnsConnsLock.RUnlock()
+		if !ok {
+			dnsConnsLock.Lock()
+			dnsConns[mapKey] = conn
+			dnsConnsLock.Unlock()
+		}
```
ppacher, line 239: There's still a race here if another connection would have the same mapKey. drop the RLock() and just Lock()

network/connection.go
```diff
@@ -0,0 +1,379 @@
...
+		// save to internal state
+		// check if it already exists
+		connsLock.RLock()
+		_, ok := conns[conn.ID]
+		connsLock.RUnlock()
+		if !ok {
+			connsLock.Lock()
+			conns[conn.ID] = conn
+			connsLock.Unlock()
+		}
```
ppacher, line 257: Same as above, I'd just Lock() and drop the RLock() instead of trying to get the lock twice

firewall/firewall.go
```diff
@@ -135,7 +137,10 @@ func handlePacket(pkt packet.Packet) {
 	// }
 
 	// allow local dns
-	if (pkt.Info().DstPort == 53 || pkt.Info().SrcPort == 53) && pkt.Info().Src.Equal(pkt.Info().Dst) {
+	if (pkt.Info().DstPort == 53 || pkt.Info().SrcPort == 53) &&
+		(pkt.Info().Src.Equal(pkt.Info().Dst) || // Windows redirects back to same interface
+			pkt.Info().Src.Equal(localhost4) || // Linux sometimes does 127.0.0.1->127.0.0.53
+			pkt.Info().Dst.Equal(localhost4)) {
```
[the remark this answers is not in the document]
dhaavi: This is meant to match both directions. When both 127.0.0.1 and 127.0.0.x are used, we'll have to match both Src and Dst.

firewall/firewall.go
```diff
@@ -135,7 +137,10 @@ func handlePacket(pkt packet.Packet) {
 	// }
 
 	// allow local dns
-	if (pkt.Info().DstPort == 53 || pkt.Info().SrcPort == 53) && pkt.Info().Src.Equal(pkt.Info().Dst) {
+	if (pkt.Info().DstPort == 53 || pkt.Info().SrcPort == 53) &&
+		(pkt.Info().Src.Equal(pkt.Info().Dst) || // Windows redirects back to same interface
```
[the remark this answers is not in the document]
dhaavi: Good one.
Although, shouldn't the compiler optimize accessing `pkg.Info()` often?

firewall/inspection/inspection.go
```diff
@@ -12,12 +12,12 @@ const (
...
 	DROP_PACKET
-	BLOCK_LINK
-	DROP_LINK
+	BLOCK_CONN
+	DROP_CONN
 	STOP_INSPECTING
 )
 
-type inspectorFn func(packet.Packet, *network.Link) uint8
+type inspectorFn func(*network.Connection, packet.Packet) uint8
```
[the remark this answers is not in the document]
dhaavi: Yup. The inspection package needs a revamp anyway, though.

firewall/inspection/inspection.go
```diff
@@ -38,20 +38,20 @@ func RegisterInspector(name string, inspector inspectorFn, inspectVerdict networ
 }
 
 // RunInspectors runs all the applicable inspectors on the given packet.
-func RunInspectors(pkt packet.Packet, link *network.Link) (network.Verdict, bool) {
+func RunInspectors(conn *network.Connection, pkt packet.Packet) (network.Verdict, bool) {
 	// inspectorsLock.Lock()
 	// defer inspectorsLock.Unlock()
```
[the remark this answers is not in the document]
dhaavi: Inspection is currently not used. I just had to replace breaking code. Will be revamped together with the encryption detection.

firewall/firewall.go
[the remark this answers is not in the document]
ppacher: Yeah but Source can only (normally) be 127.0.0.1 because that's the address assigned to the loopback interface. Destination may be something in 127.0.0.0/8 but if src **is 127.0.0.1** destination **MUST be in 127.0.0.0/8** there.

firewall/firewall.go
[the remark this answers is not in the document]
ppacher: Not sure it can because you're returning a pointer in `pkg.Info()` so he must always de-reference the pointer and cannot use a local copy on the stack/cache because the actual struct may be modified concurrently.

firewall/inspection/inspection.go
```diff
@@ -62,12 +62,12 @@ func RunInspectors(pkt packet.Packet, link *network.Link) (network.Verdict, bool
 		if skip {
 			continue
 		}
-		if link.Verdict > inspectVerdicts[key] {
+		if conn.Verdict > inspectVerdicts[key] {
 			activeInspectors[key] = true
 			continue
 		}
```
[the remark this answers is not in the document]
dhaavi: Will do.

firewall/master.go
```diff
@@ -31,91 +31,170 @@ import (
...
 		return
 	}
 
-	// grant self
-	if comm.Process().Pid == os.Getpid() {
-		log.Infof("firewall: granting own communication %s", comm)
-		comm.Accept("")
-		return
+	// check if process is communicating with itself
+	if pkt != nil {
```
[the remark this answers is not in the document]
dhaavi: Sorry, I forgot to update the docs.
They still talk about `DecideOnCommunication`, which is hard to distinguish ;)
Will update the docs.

firewall/master.go
dhaavi: There is a `*Process` instance that is used for when the process could not be identified - an "Unknown Process", which uses the Pid `-1`.

In the case of different localhost addresses this check would fail. I will add a TODO to investigate this later. I think the probability of this happening is very low though.

network/connection.go
dhaavi: Yes, that is true. One lock is better.
Also, I'm not sure if there is a performance benefit here, or if it is faster to just write to the map every time.

network/connection.go
dhaavi: See comment above.

firewall/master.go
dhaavi: Changed to Deny to Drop.

firewall/master.go
dhaavi: For DNS Queries, yes, as we don't know in which scopes the IPs of the response will be. Additional filtering is done by the `FilterDNSResponse` fn, which then looks at these separately.

firewall/master.go
dhaavi: Currently this is called everytime. I removed the second part of the comment.

firewall/master.go
dhaavi: Add logging.

firewall/inspection/inspection.go
```diff
@@ -38,20 +38,20 @@ func RegisterInspector(name string, inspector inspectorFn, inspectVerdict networ
...
+		conn.SetActiveInspectors(activeInspectors)
 	}
 
-	inspectorData := link.GetInspectorData()
+	inspectorData := conn.GetInspectorData()
 	if inspectorData == nil {
 		inspectorData = make(map[uint8]interface{})
-		link.SetInspectorData(inspectorData)
+		conn.SetInspectorData(inspectorData)
 	}
```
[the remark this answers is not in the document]
dhaavi: Will be revamped.

ppacher:
LGTM

PR #29, opened 2020-04-09
Fix updating the global profile and other bugs

dhaavi:
Also, set the blocking inbound connections to all security levels - ie. by default incoming connections are always dropped. We can reconsider this in the future, but we don't want to be opening up something without the user knowing.

### DIFF
--- a/profile/config.go
+++ b/profile/config.go
@@ -211,7 +211,7 @@ Examples:
 	if err != nil {
 		return err
 	}
-	cfgOptionBlockInbound = config.Concurrent.GetAsInt(CfgOptionBlockInboundKey, 6)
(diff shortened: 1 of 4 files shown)


profile/config.go
```diff
@@ -211,7 +211,7 @@ Examples:
 	if err != nil {
 		return err
 	}
-	cfgOptionBlockInbound = config.Concurrent.GetAsInt(CfgOptionBlockInboundKey, 6)
+	cfgOptionBlockInbound = config.Concurrent.GetAsInt(CfgOptionBlockInboundKey, 4)
```
ppacher, line 214: May add a comment why `4` or use a constant

dhaavi: Good Point, will update all of them!

PR #33, opened 2020-04-14
Add VerdictFailed and check for unknown process in nameserver

ppacher:
This PR fixes a nil pointer dereference in nameserver.go and adds a new `VerdictFailed`. It serves as the base for safing/portmaster-ui#44 which adds the new verdict to the monitor module.

### DIFF
--- a/nameserver/nameserver.go
+++ b/nameserver/nameserver.go
@@ -176,6 +176,18 @@ func handleRequest(ctx context.Context, w dns.ResponseWriter, query *dns.Msg) er
 	// get connection
 	conn := network.NewConnectionFromDNSRequest(ctx, q.FQDN, remoteAddr.IP, uint16(remoteAddr.Port))
 
+	if conn.Process() == nil {
+		tracer.Infof("nameserver: failed to find process for request %s, returning NXDOMAIN", conn)
+		returnNXDomain(w, query)
+		return nil
+	}
+
+	if conn.Process().Profile() == nil {
+		tracer.Infof("nameserver: process %s does not have a profile associated, returning NXDOMAIN", conn.Process())
+		returnNXDomain(w, query)
+		return nil
+	}
+
 	// save security level to query
 	q.SecurityLevel = conn.Process().Profile().SecurityLevel()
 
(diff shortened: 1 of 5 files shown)


nameserver/nameserver.go
```diff
@@ -176,6 +179,35 @@ func handleRequest(ctx context.Context, w dns.ResponseWriter, query *dns.Msg) er
...
+			log.Warningf("nameserver: unexpected verdict %s for connection %s, not saving", conn.Verdict, conn)
+		}
+	}()
+
+	if conn.Process().Profile() == nil {
+		tracer.Infof("nameserver: failed to find process for request %s, returning NXDOMAIN", conn)
+		returnNXDomain(w, query)
+		// FIXME(ppacher): if we save the connection (by marking it as failed)
+		// we might collect A LOT of connections for the UI.
+		//conn.Failed("Unknown process")
```
dhaavi, line 207: Did you test this? I don't think there should be that many connections. Anyway, it would be benefitial for the user to know if a process could not be correctly identified.

network/connection.go
```diff
@@ -198,6 +198,16 @@ func (conn *Connection) Deny(reason string) {
...
 
+// Failed marks the connection with VerdictFailed and stores the reason.
+func (conn *Connection) Failed(reason string) {
+	if conn.SetVerdict(VerdictFailed) {
+		conn.Reason = reason
+		log.Infof("filter: dropping connection %s because of an internal error: %s", conn, reason)
+	} else {
+		log.Warningf("filter: tried to drop %s due to error but current verdict is %s", conn, conn.Verdict)
+	}
+}
```
dhaavi, line 209: `conn.SetVerdict` will always succeed here, as `VerdictFailed` is the highest numerical verdict. This is good and intended, because we definitely want to drop if we are unable to find a Verdict without error.
We can leave this here, as you wish.

nameserver/nameserver.go
ppacher: Yes that's why I commented it out again, I went up to ~400 unknown process connections in about 2 minutes.

network/connection.go
ppacher: Ah interesting, didn't check how `SetVerdict` actually works. But yeah, I guess we can leave it in there so any future verdict (with a highter number) won't cause problems here.

nameserver/nameserver.go
dhaavi: Ok. I'll look at this later.

PR #37, opened 2020-04-17
Add support for firefox canary domain for disabling DoH

ppacher:
This PR adds support to block the Mozilla canary domain `use-application-dns.net`. When blocked, firefox does not use DoH by default and instead falls back to normal (insecure) DNS. This is required so the Portmaster can pickup the DNS requests for firewalling and forward it in a secure way.

### DIFF
--- a/profile/config.go
+++ b/profile/config.go
@@ -53,6 +53,9 @@ var (
 
 	CfgOptionRemoveBlockedDNSKey = "filter/removeBlockedDNS"
 	cfgOptionRemoveBlockedDNS    config.IntOption // security level option
+
+	CfgOptionBypassProtectionKey = "filter/preventBypassing"
+	cfgOptionBypassProtection    config.IntOption // security level option
 )
 
 func registerConfiguration() error {
@@ -324,6 +327,23 @@ Examples:
 	}
 	cfgOptionRemoveBlockedDNS = config.Concurrent.GetAsInt(CfgOptionRemoveBlockedDNSKey, int64(status.SecurityLevelsAll))
 	cfgIntOptions[CfgOptionRemoveBlockedDNSKey] = cfgOptionRemoveBlockedDNS
+
+	err = config.Register(&config.Option{
+		Name:            "Prevent Bypassing",
+		Key:             CfgOptionBypassProtectionKey,
+		Description:     "Prevent apps from bypassing the privacy filter:\n- Firefox: Disable DNS-over-HTTPs",
+		OptType:         config.OptTypeInt,
+		ExpertiseLevel:  config.ExpertiseLevelUser,
+		ReleaseLevel:    config.ReleaseLevelBeta,
+		ExternalOptType: "security level",
+		DefaultValue:    status.SecurityLevelsAll,
+		ValidationRegex: "^(7|6|4|0)",
+	})
+	if err != nil {
+		return err
+	}
+	cfgOptionBypassProtection = config.Concurrent.GetAsInt((CfgOptionBypassProtectionKey), int64(status.SecurityLevelsAll))
+	cfgIntOptions[CfgOptionBypassProtectionKey] = cfgOptionBypassProtection
 
 	return nil
 }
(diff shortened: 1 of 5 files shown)


profile/config.go
```diff
@@ -325,5 +328,22 @@ Examples:
 	cfgOptionRemoveBlockedDNS = config.Concurrent.GetAsInt(CfgOptionRemoveBlockedDNSKey, int64(status.SecurityLevelsAll))
 	cfgIntOptions[CfgOptionRemoveBlockedDNSKey] = cfgOptionRemoveBlockedDNS
 
+	err = config.Register(&config.Option{
+		Name:            "Prevent Bypassing",
+		Key:             CfgOptionBypassProtectionKey,
+		Description:     "Prevent apps from bypassing the privacy filter:\n- Firefox: Disable DNS-over-HTTPs",
```
dhaavi, line 334: I know I wrote that description, but I'm not sure this will render correctly, can you check?

profile/profile-layered.go
```diff
@@ -98,6 +100,10 @@ func NewLayeredProfile(localProfile *Profile) *LayeredProfile {
 		CfgOptionFilterSubDomainsKey,
 		cfgOptionFilterSubDomains,
 	)
+	new.PreventBypassing = new.wrapSecurityLevelOption(
+		CfgOptionBypassProtectionKey,
+		cfgOptionBypassProtection,
+	)
```
dhaavi, line 106: Please use the same wording everywhere to reduce confusion. Please stick with `PreventBypassing`.

profile/profile-layered.go
```diff
@@ -253,6 +259,22 @@ func (lp *LayeredProfile) MatchFilterLists(entity *intel.Entity) (result endpoin
...
+// MatchBypassProtection checks if the entity should be denied or permitted
+// based on some bypass protection checks.
+func (lp *LayeredProfile) MatchBypassProtection(entity *intel.Entity) (endpoints.EPResult, string) {
+	if !lp.PreventBypassing() {
+		return endpoints.NoMatch, ""
+	}
+
+	// Block firefox canary domain to disable DoH
+	if strings.ToLower(entity.Domain) == "use-application-dns.net." {
+		log.Warningf("bypass protection for firefox canary")
```
dhaavi, line 271: Was this meant for debugging?

profile/profile-layered.go
```diff
@@ -253,6 +259,22 @@ func (lp *LayeredProfile) MatchFilterLists(entity *intel.Entity) (result endpoin
...
+		return endpoints.Denied, "Firefox canary domain"
```
dhaavi, line 272: Would suggest to add more context: `blocked canary domain to prevent enabling DNS-over-HTTPS`.
`Firefox` would be in both the logs and UI anyway, plus this domain might be used by other software in the future.

profile/profile-layered.go
```diff
@@ -253,6 +259,22 @@ func (lp *LayeredProfile) MatchFilterLists(entity *intel.Entity) (result endpoin
...
+	}
+
+	return endpoints.NoMatch, ""
+}
+
```
dhaavi, line 277: I don't feel like the profiles package is the correct place to put this. How about a new file `firewall/bypassing.go` with a function similar to `DecideOnConnection` that does all the complicated prevention stuff.
This file will probably become quite a mess with all the workarounds that we will have to do.

profile/config.go
```diff
@@ -325,5 +328,22 @@ Examples:
...
+		OptType:         config.OptTypeInt,
+		ExpertiseLevel:  config.ExpertiseLevelUser,
+		ReleaseLevel:    config.ReleaseLevelBeta,
+		ExternalOptType: "security level",
+		DefaultValue:    status.SecurityLevelsAll,
+		ValidationRegex: "^(7|6|4|0)",
```
dhaavi, line 340: Is this a setting that makes sense to have turned off in the `Extreme` security level?

profile/config.go
ppacher: Nope, I just removed the "newline"

PR #38, opened 2020-04-17
Add CNAME blocking and upstream DNS resolver block detection support

ppacher:
This PR adds support to block DNS requests and connection attempts based on resolved CNAMEs. This is supported by the filterlists as well as the endpoint white/blacklist module.

In addition, this PR adds support to detect upstream DNS resolver blocking when configured.


intel/entity.go
```diff
@@ -220,10 +254,23 @@ func (e *Entity) getDomainLists() {
...
-			domains = splitDomain(domain)
-			log.Tracef("intel: subdomain list resolving is enabled, checking %v", domains)
+			for _, domain := range domainsToInspect {
+				subdomains := splitDomain(domain)
+				domains = append(domains, subdomains...)
+
+				log.Tracef("intel: subdomain list resolving is enabled: %s => %v", domains, subdomains)
+			}
+		} else {
+			domains = domainsToInspect
```
dhaavi, line 273: We could end up with really long list of domains. Would it make sense to deduplicate the domain list first? With CNAMEs and subdomains the chances are high that most are within the same suffix.

nameserver/nameserver.go
```diff
@@ -250,41 +261,32 @@ func handleRequest(ctx context.Context, w dns.ResponseWriter, query *dns.Msg) er
...
+			returnNXDomain(w, query, conn.Reason)
+			return nil
+		}
+
+		if result == endpoints.NoMatch {
+			result, reason = conn.Process().Profile().MatchFilterLists(conn.Entity)
+			if result == endpoints.Denied {
+				conn.Block("endpoint in filterlists: " + reason)
+				returnNXDomain(w, query, conn.Reason)
+				return nil
```
dhaavi, line 289: 1) Shouldn't `updateIPsAndCNAMEs` come after the CNAME filtering? Because if we block the request, the process in question should not try to access these domains via the IPs.
2) Which brings me to the question if it would make sense to move the CNAME filtering to `firewall.FilterDNSResponse` where the other filtering logic is. Would be nice to have everything in one place. (I know this was a difficult and time constrained work here)

network/connection.go
```diff
@@ -103,7 +104,7 @@ func NewConnectionFromFirstPacket(pkt packet.Packet) *Connection {
 			IP:       pkt.Info().Src,
 			Protocol: uint8(pkt.Info().Protocol),
 			Port:     pkt.Info().SrcPort,
-		}).Init()
```
dhaavi, line 106: I don't understand why you removed this, please explain. Was there a change to the `Init()` method I was not aware of? Multiple occurences.

profile/config.go
```diff
@@ -180,6 +183,23 @@ Examples:
...
+	// Include CNAMEs
+	err = config.Register(&config.Option{
+		Name:            "Filter CNAMEs",
+		Key:             CfgOptionFilterCNAMEKey,
+		Description:     "Also filter requests where a CNAME would be blocked",
+		OptType:         config.OptTypeInt,
+		ExternalOptType: "security level",
+		DefaultValue:    status.SecurityLevelsAll,
+		ValidationRegex: "^(7|6|4)$",
+	})
```
dhaavi, line 195: CNAMEs are a very technical thing, would suggest assigning to `ExperienceLevelExpert`.

nameserver/nameserver.go
```diff
@@ -240,8 +245,14 @@ func handleRequest(ctx context.Context, w dns.ResponseWriter, query *dns.Msg) er
 	if err != nil {
 		// TODO: analyze nxdomain requests, malware could be trying DGA-domains
 		tracer.Warningf("nameserver: %s requested %s%s: %s", conn.Process(), q.FQDN, q.QType, err)
-		returnNXDomain(w, query)
-		conn.Failed("failed to resolve: " + err.Error())
+
+		if _, ok := err.(*resolver.BlockedUpstreamError); ok {
```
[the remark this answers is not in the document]
ppacher: Yeah probably, I'm still used to use a type check ...

intel/entity.go
ppacher: Good point, didn't thought about that

nameserver/nameserver.go
ppacher: > 1. Shouldn't updateIPsAndCNAMEs come after the CNAME filtering? Because if we block the request, the process in question should not try to access these domains via the IPs.

updateIPsAndCNAMEs aggregates the CNAMEs that were used to resolve the requested domain to the IP and updates the connections Entity as well as the `IPInfo` (or creates it). That aggregation and update must happen before the CNAME filtering. We coulde move the `IPInfo` update after the CNAME filtering but I'm not sure if there's a benefit. If we block the request the process wouldn't receive the answer and thus don't know the IP to connect to. Or did I get something wrong here?

> 2. Which brings me to the question if it would make sense to move the CNAME filtering to firewall.FilterDNSResponse where the other filtering logic is. Would be nice to have everything in one place. (I know this was a difficult and time constrained work here)

I'm not sure, we don't actually "filter out" CNAME resource records form the DNS anwser section like `FilterDNSResponse` does. We just check if one of the CNAMEs is forbidden and deny the whole request here. `FilterDNSResponse` removes dedicated RR entries from the response and only causes the response to be dropped/blocked if not IPs are left over. 
For simplicity we could/should move that rechecking into a dedicated function but IMHO it doesn't fit into `FilterDNSResponse`.

network/connection.go
ppacher: That method does nothing now and I actually wanted to completely get rid of it but I just removed this one line :facepalm: 

https://github.com/safing/portmaster/blob/e82ec22f0777c5c0a1c4817cf95b312b19dc6a38/intel/entity.go#L76-L79

profile/config.go
ppacher: Good point

PR #39, opened 2020-04-17
Add support for unidentified processes and profiles

dhaavi:


process/special.go
```diff
@@ -0,0 +1,67 @@
...
+
+	// mark profile as used
+	localProfile.MarkUsed()
+
+	p.LocalProfileKey = localProfile.Key()
+	p.profile = profile.NewLayeredProfile(localProfile)
+
+	go p.Save()
+	return p
+}
```
ppacher, line 67: I don't get why we actually need this function. Isn't it enough to ensure the special profiles for Unidentified and System exist on startup (module preperation/start). Since it's not allowed to call that function for the same process more than once I'm not sure what it's actual purpose is.

profile/active.go
```diff
@@ -38,7 +45,34 @@ func markActiveProfileAsOutdated(scopedID string) {
 
 	profile, ok := activeProfiles[scopedID]
 	if ok {
-		profile.oudated.Set()
+		profile.outdated.Set()
 		delete(activeProfiles, scopedID)
 	}
 }
+
+func cleanActiveProfiles(ctx context.Context) error { //nolint:param // need to conform to interface
```
ppacher, line 53:
```suggestion
func cleanActiveProfiles(_ context.Context) error {
```

network/connection.go
```diff
@@ -41,6 +41,7 @@ type Connection struct { //nolint:maligned // TODO: fix alignment
 	VerdictPermanent bool
 	Inspecting       bool
 	Encrypted        bool // TODO
+	Hidden           bool
```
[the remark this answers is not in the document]
dhaavi: As discussed, will rename to `Internal`.

network/dns.go
```diff
@@ -23,7 +23,16 @@ func removeOpenDNSRequest(pid int, fqdn string) {
...
+	_, ok := openDNSRequests[key]
+	if ok {
+		delete(openDNSRequests, key)
+		return
+	}
+
+	// check if there is an open dns request from an unidentified process
+	if pid >= 0 {
+		delete(openDNSRequests, "-1/"+fqdn)
+	}
```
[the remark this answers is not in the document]
dhaavi: Will update as discussed.

process/special.go
dhaavi: Will update to use process caching.

PR #42, opened 2020-04-21
Add update module status, allow disabling of updates

ppacher:


updates/config.go
```diff
@@ -32,31 +35,65 @@ func registerConfig() error {
...
+		changed = true
+		forceUpdate = !updatesCurrentlyDisabled
+	}
+
 	if changed {
 		registry.SelectVersions()
 		module.TriggerEvent(VersionUpdateEvent, nil)
+
+		if forceUpdate {
+			module.Resolve(updateFailed)
```
dhaavi, line 90: Unnecessary resolve, as module errors are always replaced.

updates/config.go
```diff
@@ -32,31 +35,65 @@ func registerConfig() error {
...
+			TriggerUpdate()
```
dhaavi, line 91: lint error?

updates/main.go
```diff
@@ -156,14 +199,33 @@ func DisableUpdateSchedule() error {
...
+		return nil
+	}
+	defer log.Debugf("updates: finished checking for updates")
+
+	module.Hint(updateInProgress, updateInProcessDescr)
+
+	defer func() {
+		if err == nil {
+			module.Resolve(updateInProgress)
+			module.Resolve(updateFailed)
```
dhaavi, line 214: Unnecessary resolve, as module errors are always replaced.

PR #44, opened 2020-04-27
WIP: Pre alpha finalizing

dhaavi:
This a smorgasbord (🤣 what a word!) of commits - please review them separately.


profile/config.go
```diff
@@ -257,9 +257,10 @@ Examples:
 	err = config.Register(&config.Option{
 		Name:            "Block Scope Local",
 		Key:             CfgOptionBlockScopeLocalKey,
-		Description:     "Block connections to your own device, ie. localhost.",
+		Description:     "Block connections internally on own device, ie. localhost.",
```
ppacher, line 260:
```suggestion
		Description:     "Block connections internally on your own device, ie. localhost.",
```

or 

```suggestion
		Description:     "Block connections internally to this device, ie. localhost.",
```

profile/config.go
```diff
@@ -278,7 +279,7 @@ Examples:
 		Order:           cfgOptionBlockScopeLANOrder,
 		OptType:         config.OptTypeInt,
 		ExternalOptType: "security level",
-		DefaultValue:    status.SecurityLevelOff,
+		DefaultValue:    status.SecurityLevelsNormalAndExtreme,
```
ppacher, line 282: shouldn't that be `SecurityLevelsHighAndExtreme`? Does `SecurityLevelsNormalAndExtreme` even make sense to exist?

dhaavi: Yes, this should in fact be `SecurityLevelsHighAndExtreme`.
The Portmaster actually refused to start, because the default did not match the regex! Just did not build and run it ;)
Currently, `SecurityLevelsNormalAndExtreme` does not make sense, but exists historically.

PR #53, opened 2020-05-22
Remove version information from moved file

dhaavi:
Otherwise the file is being picked up by the update registry scan.


dhaavi:
Licensing is ok - not an error.

PR #55, opened 2020-05-25
Fix portmaster control date format

dhaavi:


dhaavi:
CI is being funny. All tests are ok.

PR #66, opened 2020-06-03
Increase resilience of api auth by waiting long and trying more often

### LINKED ISSUE #64: Improve Portmaster API authentication failure error message
**What would you like to be added**:

Return a helpful error message when authentication in the Portmaster API fails. Especially explain why the request was denied and what workarounds exists (eg. dev mode).

**Why is this needed**:

When accessing the Portmaster API from outside the portmaster ecosystem, the feedback ("Forbidden") is not helpful at all.

dhaavi:
Also, improve messages sent to user.

Closes #64.

### DIFF
--- a/firewall/api.go
+++ b/firewall/api.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/safing/portbase/api"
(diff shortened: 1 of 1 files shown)


dhaavi:
Build fails because it requires https://github.com/safing/portbase/pull/54.

PR #68, opened 2020-06-03
Remove " (deleted)" suffix on linux for deleted files

dhaavi:
Not removing the " (deleted)" suffix also leads to the creation of profiles. Deletion occurs frequently when executables are upgraded while running.


dhaavi:
CI fails on Windows because of this weird thing again. Not related to this PR.

PR #78, opened 2020-06-16
Add TCP/TLS pipelining dns resolver

dhaavi:


resolver/resolver-tcp.go
```diff
@@ -0,0 +1,359 @@
...
+	tr.dnsClient.Net = "tcp-tls"
+	tr.dnsClient.TLSConfig = &tls.Config{
+		MinVersion: tls.VersionTLS12,
+		ServerName: tr.resolver.VerifyDomain,
+		// TODO: use portbase rng
+	}
+	return tr
+}
+
+func (tr *TCPResolver) client(workerCtx context.Context) error { //nolint:gocognit,gocyclo // TODO
```
ppacher, line 89: This function is so huge I'm feeling lost while reviewing it. Seems like there are multiple things that could be extracted into dedicated functions (like re-injecting queries).  IMHO this seems over-engineered and is likely not unit-testable (or debug-able :D). To be easily understandable (and thus likely less buggy) a function should not be longer ~25 lines and don't have more than 3-4 levels of indention. Let's split up this one :)

resolver/resolver-tcp.go
```diff
@@ -0,0 +1,359 @@
...
+		}
+		// connect
+		c, err := tr.dnsClient.Dial(tr.resolver.ServerAddress)
+		if err != nil {
+			tr.ReportFailure()
+			log.Debugf("resolver: failed to connect to %s (%s)", tr.resolver.Name, tr.resolver.ServerAddress)
+			continue connMgmt
+		}
+		tr.dnsConnection = c
+		log.Debugf("resolver: connected to %s (%s)", tr.resolver.Name, tr.dnsConnection.RemoteAddr())
```
ppacher, line 188: no locking required when setting those values on `tr`?  Seems like a good candidate for a dedicated method.

PR #83, opened 2020-06-26
Fix netenv resolver interaction

dhaavi:
Also fix endless loop in tcp resolver when network is down

### DIFF
--- a/nameserver/nameserver.go
+++ b/nameserver/nameserver.go
@@ -106,17 +106,19 @@ func handleRequestAsWorker(w dns.ResponseWriter, query *dns.Msg) {
 }
 
 func handleRequest(ctx context.Context, w dns.ResponseWriter, query *dns.Msg) error { //nolint:gocognit // TODO
-	// return with server failure if offline
-	if netenv.GetOnlineStatus() == netenv.StatusOffline {
-		returnServerFailure(w, query)
-		return nil
-	}
-
 	// only process first question, that's how everyone does it.
 	question := query.Question[0]
 	q := &resolver.Query{
 		FQDN:  question.Name,
 		QType: dns.Type(question.Qtype),
+	}
+
+	// return with server failure if offline
+	if netenv.GetOnlineStatus() == netenv.StatusOffline &&
+		!netenv.IsOnlineStatusTestDomain(q.FQDN) {
+		log.Tracer(ctx).Debugf("resolver: not resolving %s, device is offline", q.FQDN)
+		returnServerFailure(w, query)
+		return nil
 	}
 
 	// check class
--- a/netenv/online-status.go
+++ b/netenv/online-status.go
@@ -106,6 +106,8 @@ var (
 
 	captivePortalURL  string
 	captivePortalLock sync.Mutex
+
+	waitForever = make(chan time.Time)
 )
 
 func init() {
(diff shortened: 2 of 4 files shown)


nameserver/nameserver.go
```diff
@@ -106,19 +106,21 @@ func handleRequestAsWorker(w dns.ResponseWriter, query *dns.Msg) {
...
 		QType: dns.Type(question.Qtype),
 	}
 
+	// return with server failure if offline
+	if netenv.GetOnlineStatus() == netenv.StatusOffline &&
+		!netenv.IsOnlineStatusTestDomain(q.FQDN) {
+		log.Tracer(ctx).Debugf("resolver: not resolving %s, device is offline", q.FQDN)
+		returnServerFailure(w, query)
+		return nil
+	}
```
ppacher, line 122: Hm, AFAIK the online-status is about "real" internet access. Shouldn't we still allow local DNS/traffic to work?

netenv/online-status.go
```diff
@@ -221,6 +225,21 @@ func monitorOnlineStatus(ctx context.Context) error {
...
+func getDynamicStatusTrigger() <-chan time.Time {
+	switch GetOnlineStatus() {
+	case StatusOffline:
+		return time.After(10 * time.Second)
+	case StatusLimited, StatusPortal:
+		return time.After(1 * time.Minute)
+	case StatusSemiOnline:
+		return time.After(5 * time.Minute)
+	case StatusOnline:
+		return waitForever
```
ppacher, line 237: just return `nil` here because receiving from a `nil` channel blocks forever

resolver/resolve.go
```diff
@@ -213,13 +219,24 @@ func deduplicateRequest(ctx context.Context, q *Query) (finishRequest func()) {
...
 
+	// check if we are online
+	if netenv.GetOnlineStatus() == netenv.StatusOffline {
+		if netenv.IsOnlineStatusTestDomain(q.FQDN) {
+			log.Tracer(ctx).Debugf("resolver: permitting online status test domain %s to resolve even though offline", q.FQDN)
+		} else {
+			log.Tracer(ctx).Debugf("resolver: not resolving %s, device is offline", q.FQDN)
+			// we are offline and this is not an online check query
+			return nil, ErrOffline
+		}
```
ppacher, line 237:
```suggestion
		if !netenv.IsOnlineStatusTestDomain(q.FQDN) {
			log.Tracer(ctx).Debugf("resolver: not resolving %s, device is offline", q.FQDN)
			// we are offline and this is not an online check query
			return nil, ErrOffline
		}
		log.Tracer(ctx).Debugf("resolver: permitting online status test domain %s to resolve even though offline", q.FQDN)
```

resolver/resolve.go
```diff
@@ -246,6 +263,11 @@ resolveLoop:
 				case errors.Is(err, ErrBlocked):
 					// some resolvers might also block
 					return nil, err
+				case netenv.GetOnlineStatus() == netenv.StatusOffline &&
+					!netenv.IsOnlineStatusTestDomain(q.FQDN):
+					log.Tracer(ctx).Debugf("resolver: not resolving %s, device is offline", q.FQDN)
+					// we are offline and this is not an online check query
+					return nil, ErrOffline
```
ppacher, line 270: May also add a `default:` and log the error so it's easier to debug why a given resolver doesn't work

PR #89, opened 2020-07-10
[WIP] Migrate to go modules

ppacher:
Migration to go modules:

- [x] Migrate from `dep` to go.mod
- [x] Update third party dependencies to their go.mod version
- [ ] Update ./test script to work with go.mod (or replace with GitHub actions and use https://github.com/nektos/act)


dhaavi:
PR is obsolete now.

PR #90, opened 2020-07-12
Implement Post Factum Review Suggestions

dhaavi:
There is one commit for each PR.


resolver/resolver-tcp.go
```diff
@@ -386,3 +145,295 @@ func (tr *TCPResolver) Query(ctx context.Context, q *Query) (*RRCache, error) {
...
+		}
+	}
+
+	return true
+}
+
+func (mgr *tcpResolverConnMgr) establishConnection() (success bool) {
+	// create connection
+	mgr.connCtx, mgr.cancelConnCtx = context.WithCancel(mgr.workerCtx)
+	mgr.connClosing = abool.New()
```
ppacher, line 297: Return the connection context and the cancel function instead of storing them in the struct (see package doc link above). `cancelConnCtx` is only used by `queryHandler` and `msgReader` which can get it as a parameter (use a closure in StartServiceWorker). Since calling a context.CancelFunc more than once is a no-op you don't need to guard the call to `cancelConnCtx` in `mgr.connClosing.SetToIf(false, true)`, just `defer cancelConnCtx()` in `queryHandler()` and remove the call in `cleanupConnections()` (because it MUST have been called in `queryHandler` anyway)

resolver/resolver-tcp.go
```diff
@@ -386,3 +145,295 @@ func (tr *TCPResolver) Query(ctx context.Context, q *Query) (*RRCache, error) {
...
+	}
+	mgr.conn = c
+	log.Debugf("resolver: connected to %s (%s)", mgr.tr.resolver.Name, mgr.conn.RemoteAddr())
+
+	// reset timer
+	mgr.connTimer.Stop()
+	select {
+	case <-mgr.connTimer.C: // try to empty the timer
+	default:
+	}
```
ppacher, line 319: [Package doc](https://golang.org/pkg/time/#Timer.Stop) suggests:

```suggestion
	if !mgr.connTimer.Stop() {
		<-mgr.connTimer.C: // try to empty the timer
	}
```

resolver/resolver-tcp.go
```diff
@@ -386,3 +145,295 @@ func (tr *TCPResolver) Query(ctx context.Context, q *Query) (*RRCache, error) {
...
+					}
+				} else {
+					log.Debugf(
+						"resolver: received possibly unsolicited reply from %s (%s): txid=%d q=%+v",
+						mgr.tr.resolver.Name,
+						mgr.conn.RemoteAddr(),
+						msg.Id,
+						msg.Question,
+					)
+				}
```
ppacher, line 404: Consider moving that to a dedicated function as it does not depend on loop vars and the possibility of returning allows to outdent multiple blocks:

```golang
func (mgr *tcpResolverConnMgr) handleQueryResponse(msg *dns.Msg) {
	// handle query from resolver
	mgr.tr.Lock()
	inFlight, ok := mgr.tr.inFlightQueries[msg.Id]
	if ok {
		delete(mgr.tr.inFlightQueries, msg.Id)
	}
	mgr.tr.Unlock()

	if !ok {
		log.Debugf(
			"resolver: received possibly unsolicited reply from %s (%s): txid=%d q=%+v",
			mgr.tr.resolver.Name,
			mgr.conn.RemoteAddr(),
			msg.Id,
			msg.Question,
		)
        	return
	}

	select {
	case inFlight.Response <- msg:
		mgr.failCnt = 0 // reset fail counter
		// responded!
	default:	
	}

	// save to cache, if enabled
	if !inFlight.Query.NoCaching {
		// persist to database
		rrCache := inFlight.MakeCacheRecord(msg)
		rrCache.Clean(600)
		err := rrCache.Save()
		if err != nil {
			log.Warningf(
				"resolver: failed to cache RR for %s%s: %s",
				inFlight.Query.FQDN,
				inFlight.Query.QType.String(),
				err,
			)
		}
	}
}
```

resolver/resolver-tcp.go
```diff
@@ -348,17 +103,21 @@ func (tr *TCPResolver) submitQuery(_ context.Context, q *Query) *InFlightQuery {
 		ConnInstanceID: atomic.LoadUint32(tr.connInstanceID),
 	}
 	tr.Lock()
```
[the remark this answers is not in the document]
dhaavi: Resolved in discussion: TCPResolver is only used internally.

resolver/resolver-tcp.go
dhaavi: Resolved in discussion: It was like in the docs before, but there was a problem with that. I currently can't recall what the problem was exactly.

dhaavi: This has now been removed anyway.

dhaavi:
Windows fail is unrelated and our borderline CI bug.

PR #91, opened 2020-07-12
Update default dns servers

dhaavi:
According to https://safing.io/blog/2020/07/07/we-are-updating-portmasters-default-dns-servers/

### DIFF
--- a/resolver/config.go
+++ b/resolver/config.go
@@ -13,46 +13,43 @@ var (
 	defaultNameServers = []string{
 		// Collection of default DNS Servers
 
-		// Default servers should be:
-		// Anycast:
-		// - Servers should be reachable from anywhere with reasonable latency.
-		// - Servers should be near to the user for geo-content to work correctly.
-		// Private:
-		// - Servers should not do any or only minimal logging.
-		// - Available logging data may not be used against the user, ie. unethically.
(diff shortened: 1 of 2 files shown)


resolver/config.go
```diff
@@ -13,46 +13,43 @@ var (
...
-		// - Available logging data may not be used against the user, ie. unethically.
-
-		// Sadly, only a few services come close to fulfilling these requirements.
-		// For now, we have settled for two bigger and well known services: Quad9 and Cloudflare.
-		// TODO: monitor situation and re-evaluate when new services become available
-		// TODO: explore other methods of making queries more private
+		// For a detailed explanation how we choose our default resolvers, check out
+		// https://safing.io/blog/2020/07/07/how-safing-selects-its-default-dns-providers/
+
+		// These resolvers define a working set. Which provider we selected as the
```
davegson, line 19: for future notice, for more clarity I'd suggest:

> These resolvers define a working set. Which provider we select as the primary and which providers are (in)active backups depends on the current situation.

And then describe each provider with an additional line:

```
// Quad9 (encrypted DNS)
// inactive backup
...

// Cloudflare (encrypted DNS, with malware protection)
// primary default
..

// AdGuard (encrypted DNS, default flavor)
// active backup

```

PR #94, opened 2020-07-15
Improve captive portal handling

dhaavi:


northys:
I'm not sure if this is part of this PR but today it is first time when I got notification `Portmaster detected captive portal at cdwifi.cz` after connecting to wifi @ ČD train.

Issue #77, opened 2020-06-10
Add mechanisms to clear the DNS cache to portmasterctl


dhaavi:
**What would you like to be added**:

The DNS cache should be clearable, either in full or for specific records. This could be achieved using the following mechanisms, which should be understood as suggestions:
- If there are three consecutive requests (without any other in the middle) for the same domain, clear the cache for that domain and make a new request.
- If a query for `orf.at.clear.portmaster.` is made, the dns cache for `orf.at.` with the same query type is cleared.
- If a query for `clear.portmaster.` is made, the complete dns cache is cleared.
- If a query for `orf.at.fresh.portmaster` is made, the dns cache for `orf.at.` with the same query type is cleared, a fresh query is run and the results returned (unchanged?).

Before this is implemented, we should also think about potential attacks this could enable. For example, an attacker could circumvent being blocked by just resolving using `bad.example.com.fresh.portmaster.`. Also, it would be easy to determine if a user is running the Portmaster or not.

**Why is this needed**:

Developers and system administration often have to wait for updated DNS records to arrive at their machine. This would let them speed up the process.

dhaavi:
Well, with all kinds of DNS related changes, admins usually monitor the change, especially when it is blocking their work. In the case you described, you were waiting until the DNS change finished, in order to run an ACME client. Running fresh queries would improve the accuracy of this approach. Especially with something as an ACME server, which is very unlikely to have a cached record of the domain in question, or even use a cache at all, all you want to know if the authoritative nameserver has been updated. Bypassing as many caches as possible is therefore desirable.

> Or is there any other DNS client that uses DNS query to purge cache?

No, not that I know of.

> What is the difference between `fresh.portmaster.` and `clear.portmaster.`. The `fresh.portmaster.` one returns fresh data returned by new query and `clear.portmaster.` returns what? Cached result being removed? Empty response?

A query for `clear.portmaster.` would probably return a TXT record with `Portmaster DNS cache cleared`.
 
> Regarding security issues - wouldn't it be better to take advantage of the same authorization process which is used by portmaster dashboard to authenticate to the portmaster core API? This could solve the security issues you have mentioned.

Authentication to the Portmaster Core API just checks if the process is part of the Portmaster. That would not work in this case, as the querying process would always be a process that isn't part of the Portmaster.
 
[...]

PR #96, opened 2020-07-16
Refactor pmctl to new portmaster-start

### LINKED ISSUE #70: Stop upgrading portmaster-control in $PATH on linux
_This is follow-up issue for https://github.com/safing/portmaster-packaging/issues/1_

Also, check if portmaster-control is in path and remove for a couple versions.

Follow-up issue: create new issue to remove portmaster-control purging.

ppacher:
This PR refactors and replaces the old pmctl/portmaster-control with the new portmaster-start binary. It's now located under `cmds/portmaster-start` and includes a few breaking changes and fixes.

The most important ones are:

**Already deprecated flags dropped**

Support for `--db` has finally be dropped. All users must migrate to `--data`.

**`PORTMASTER_DATA` environment variable fixed**

Due to how the legacy portmaster-control handled command arguments the `PORTMASTER_DATA` environment variable never worked. Now it does! So instead of using `--data /var/lib/portmaster` all the time a simple `export PORTMASTER_DATA=/var/lib/portmaster` should be enough. One can verify that using the built-in `show` command.

**The `run` sub-command has been removed**

The old `run` sub-command has been removed and it's functionality has been integrated into the root command. That is, instead of `./portmaster-start run core` you just need to use `./portmaster-start core` (which doesn't stutter). 

**Additional parameters for components**

Previously all arguments after the third one have been passed when starting portmaster components. That caused weird things like `portmaster-control --data=/path run notifier` to break because instead of passing `--data=xxx` down to the notifier it actually passed the string `notifier` (wrong position of `--data`). In addition, the legacy portmaster-control did not perform any flag validation which means a typo like `--dat` instead of `--data` did not cause an error. 

The new portmaster-start does more strict parameter and flag validation but comes at the cost that parameters that need to be passed down to the component itself must be separated with `--`. For example:

```bash
# Previously with legacy portmaster-control (note the required position of --data)
./portmaster-control run  core --data=/tmp --log-level debug

# Should now be written like this:
./portmaster-start --data=/tmp core -- --log-level debug
```

That also allows passing `-h` or `--help` down to the underlying component:
```
portmaster-start --max-retries 0 --data /tmp core -- -h
```

**`--version` moved to dedicated sub-command**

The old `--version` and `--ver` flags have been replaced by a new `version` sub-command:

```bash
# previously
portmaster-control --version
# now:
portmaster-start version
```

```bash
# previously:
portmaster-control --ver
# now:
portmaster-start version --short
```

In addition, there's now a `--all` flag for the `version` sub-command that displays versions for all available resources. If combined with `--short` only resources that matter for that platform (i.e. now windows binaries on linux) are shown. Note that `--all` requires `--data` to be set.

```bash
/portmaster-start version --all --short --data /opt/portmaster
portmaster-start 0.3.5

Components:
   all/intel/geoip/geoip-asn.mmdb                  2019.13.08
   linux_amd64/notifier/portmaster-notifier        0.1.8
   linux_amd64/packages/portmaster-installer.deb   0.0.1
#
# ...
#
```

**New `recover-iptables` sub-command`**

There's also a new sub-command `recover-iptables` that can be used to cleanup the iptables rules after an unclean shutdown of portmaster. See #6 for more information.

**Additional changes**

- portmaster binary is now located under `cmds/portmaster-core`
- imported uptool from portbase to `cmds/uptool`
- add default implementation for interception module so portmaster can compile on other platforms as well

--

**Related issues**

- The reason of changing `portmaster-control` to `portmaster-start` follows the discussion in https://github.com/safing/portmaster/issues/77.
- Due to the rename this PR also fixes #70 . Instead of upgrading it a warning is logged if it's not the new portmaster-start binary.


cmds/portmaster-start/shutdown.go
```diff
@@ -5,25 +5,35 @@ import (
...
+	select {
+	case <-shuttingDown:
+		return
+	default:
 		shutdownError = err
 		close(shuttingDown)
 	}
 }
 
+func isShutdown() bool {
```
dhaavi, line 28: A bit weird.

cmds/portmaster-start/run.go
```diff
@@ -0,0 +1,384 @@
...
+			AllowHidingWindow: false,
+		},
+		{
+			Name:              "Portmaster Notifier",
+			Identifier:        "notifier/portmaster-notifier",
+			AllowDownload:     false,
+			AllowHidingWindow: true,
+		},
+	})
+}
```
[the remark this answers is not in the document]
dhaavi: Found it! :)

cmds/portmaster-start/update.go
```diff
@@ -26,15 +26,15 @@ func downloadUpdates() error {
...
 			platform("notifier/portmaster-snoretoast.exe"),
 		}
 	} else {
 		registry.MandatoryUpdates = []string{
 			platform("core/portmaster-core"),
-			platform("control/portmaster-control"),
+			platform("start/portmaster-start"),
 			platform("app/portmaster-app"),
 			platform("notifier/portmaster-notifier"),
 		}
```
dhaavi, line 40: Please also add the base ui module.

updates/upgrader.go
```diff
@@ -113,15 +113,15 @@ func upgradePortmasterControl() error {
...
+	err := upgradeFile(rootPmStartPath, pmCtrlUpdate)
 	if err != nil {
 		return err
 	}
-	log.Infof("updates: upgraded %s", rootControlPath)
+	log.Infof("updates: upgraded %s", rootPmStartPath)
 
-	// upgrade parent process, if it's portmaster-control
+	// upgrade parent process, if it's portmaster-start
 	parent, err := processInfo.NewProcess(int32(os.Getppid()))
```
dhaavi, line 125: Do we want to continue to upgrade the parent?

cmds/portmaster-start/main.go
```diff
@@ -0,0 +1,220 @@
...
+		dataDir = os.Getenv("PORTMASTER_DATA")
+	}
+
+	// check data dir
+	if dataDir == "" {
+		return errors.New("please set the data directory using --data=/path/to/data/dir")
+	}
+
+	// remove redundant escape characters and quotes
+	dataDir = strings.Trim(dataDir, `\"`)
```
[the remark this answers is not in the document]
dhaavi: It does something different. Did not get that.

cmds/portmaster-start/main.go
```diff
@@ -0,0 +1,220 @@
...
+func updateRegistryIndex() {
+	err := registry.LoadIndexes()
+	if err != nil {
+		log.Printf("WARNING: error loading indexes: %s\n", err)
+	}
+
+	err = registry.ScanStorage("")
+	if err != nil {
+		log.Printf("WARNING: error during storage scan: %s\n", err)
+	}
```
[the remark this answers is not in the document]
dhaavi: Resolved: Error does not mean it failed. It could also only partially fail and still work.

cmds/uptool/main.go
```diff
@@ -0,0 +1,34 @@
...
+	Short: "A simple tool to assist in the update and release process",
+	Args:  cobra.ExactArgs(1),
+	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+		absPath, err := filepath.Abs(args[0])
+		if err != nil {
+			return err
+		}
+
+		registry = &updater.ResourceRegistry{}
+		return registry.Initialize(utils.NewDirStructure(absPath, 0o755))
```
dhaavi, line 25: If I understand correctly, these permissions are the same as 0755. I like the explicit octal typing. Maybe we should adopt this everywhere.
It'd be nice if the Go stdlib also did this.

updates/upgrader.go
```diff
@@ -113,38 +113,62 @@ func upgradePortmasterControl() error {
...
 
-	return nil
+	root := filepath.Dir(registry.StorageDir().Path)
+	if !strings.HasPrefix(absPath, root) {
+		log.Warningf("detected unexpected path %s for portmaster-start", absPath)
+
+		// TODO(ppacher): once we released a new installer and folks had time
+		//                to update we should send a module warning/hint to the
+		//                UI notifying the user that he's using portmaster-start
+		//                from a wrong location.
```
dhaavi, line 170: Can this happen at all if we make the switch to portmaster-start?

I'd rather not update it at all outside of the data dir and just give a warning to the user whenever the parent _is_ portmaster-start, but _is not_ in the data directory - forever.

dhaavi: Got something mixed up. Nevermind.

ppacher:
Merging blocked until Monday 20.07.2020

ppacher:
PR is now rebased on latest develop branch.

updates/upgrader.go
```diff
@@ -140,38 +140,62 @@ func upgradePortmasterControl() error {
...
```
dhaavi, line 197: As discussed earlier, we can actually do this immediately.

updates/upgrader.go
```diff
@@ -140,38 +140,62 @@ func upgradePortmasterControl() error {
...
+	rootPmStartPath := filepath.Join(filepath.Dir(registry.StorageDir().Path), filename)
+	err := upgradeFile(rootPmStartPath, pmCtrlUpdate)
 	if err != nil {
 		return err
 	}
-	log.Infof("updates: upgraded %s", rootControlPath)
+	log.Infof("updates: upgraded %s", rootPmStartPath)
 
-	// upgrade parent process, if it's portmaster-control
+	// TODO(ppacher): remove once we released a few more versions ...
```
dhaavi, line 151: This can actually stay forever.

cmds/portmaster-start/version.go
```diff
@@ -0,0 +1,79 @@
...
+				return err
+			}
+		}
+
+		return nil
+	},
+	RunE: func(*cobra.Command, []string) error {
+		if !showAllVersions {
+			if showShortVersion {
+				fmt.Println(info.Version())
```
dhaavi, line 35: Shouldn't we `return nil` here?

cmds/portmaster-start/version.go
```diff
@@ -0,0 +1,79 @@
...
+
+		return nil
+	},
+}
+
+func init() {
+	flags := versionCmd.Flags()
+	{
+		flags.BoolVar(&showShortVersion, "short", false, "Print only the version number.")
+		flags.BoolVar(&showAllVersions, "all", false, "Dump versions for all components.")
```
dhaavi, line 75: Would suggest using `assets` instead of `components`, as we use `components in the UI for something different. This might cause confusion.

cmds/portmaster-start/version.go
ppacher: Yes, we should!

PR #107, opened 2020-07-22
Enable building on unsupported systems

dhaavi:


process/process_linux.go
```diff
@@ -7,7 +7,7 @@ func (p *Process) IsUser() bool {
 
 // IsAdmin returns whether the process is run by an admin user.
 func (p *Process) IsAdmin() bool {
-	return p.UserID >= 0
+	return p.UserID >= 0 && p.UserID < 1000
 }
```
ppacher, line 11: May remove that check completely

dhaavi:
PR was approved with change.

Issue #110, opened 2020-07-27
Silent NFQueue Fail


dhaavi:
Creating a new issue for something that came up in another thread:

---

The network has stopped working in bar :(  `systemctl restart portmaster` fixed it as usually.

Sadly I have recently changed log level to info..

```
čec 25 23:01:37 northys-laptop portmaster-control[42451]: 200725 23:01:37.682 connection:207 ▶ INFO 076 filter: granting connection Unknown::-1 to maps.gstatic.com. (172.217.23.195), endpoint is not blacklisted (default=permit)
čec 25 23:01:37 northys-laptop portmaster-control[42451]: 200725 23:01:37.775 connection:207 ▶ INFO 077 filter: granting connection Unknown::-1 to maps.gstatic.com. (172.217.23.195), endpoint is not blacklisted (default=permit)
čec 25 23:01:37 northys-laptop portmaster-control[42451]: 200725 23:01:37.867 connection:207 ▶ INFO 078 filter: granting connection Unknown::-1 to maps.gstatic.com. (172.217.23.195), endpoint is not blacklisted (default=permit)
čec 25 23:01:49 northys-laptop portmaster-control[42451]: 200725 23:01:49.145 er/resolve:204 ▶ WARN 079 resolver: async query for www.gstatic.com.A failed: all 3 query-compliant resolvers failed, last error: read udp 192.168.101.52:58825->213.46.172.37:53: i/o timeout
čec 25 23:01:53 northys-laptop portmaster-control[42451]: 200725 23:01:53.983 tel/entity:170 ▶ WARN 080 intel: failed to resolve IP 192.168.101.42: failed to resolve 42.101.168.192.in-addr.arpa.PTR: all 1 query-compliant resolvers failed, last error: failed to send query: write udp6 [::]:44732->[ff02::fb]:5353: sendto: cannot assign requested address
čec 25 23:01:53 northys-laptop portmaster-control[42451]: 200725 23:01:53.983 connection:235 ▶ INFO 081 filter: dropping connection Unknown::-1 <- 192.168.101.42, endpoint is not whitelisted (incoming is always default=block)
[...]

PR #112, opened 2020-07-30
Add experimental nfqueue interception backend

ppacher:
This PR adds a new experimental nfqueue interception backed based on [florianl/go-nfqueue](https://github.com/florianl/go-nfqueue) (MIT licensed). It should add more stability to the interception of network packets and also eliminates the CGO and the runtime dependency on libnetfilter_queue as it uses a AF_NETLINK socket directly. Note that this commit does not remove the old nfqueue backend yet but adds a `--experimental-nfqueue` flag to portmaster-core until it eventually becomes the default.

### DIFF
--- a/firewall/interception/nfqueue_linux.go
+++ b/firewall/interception/nfqueue_linux.go
@@ -1,13 +1,17 @@
 package interception
 
 import (
+	"flag"
 	"fmt"
 	"sort"
 	"strings"
 
 	"github.com/coreos/go-iptables/iptables"
 
+	"github.com/safing/portbase/log"
+	"github.com/safing/portmaster/firewall/interception/nfqexp"
 	"github.com/safing/portmaster/firewall/interception/nfqueue"
+	"github.com/safing/portmaster/network/packet"
 )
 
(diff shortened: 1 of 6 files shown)


ppacher:
This change may also fix #110

firewall/interception/nfqueue_linux.go
```diff
@@ -203,32 +222,42 @@ func deactivateIPTables(protocol iptables.Protocol, rules, chains []string) erro
...
 	}
-	out6Queue, err = nfqueue.NewNFQueue(17060)
+	out6Queue, err = nfQueueFactory(17060, true)
 	if err != nil {
 		_ = Stop()
-		return fmt.Errorf("interception: failed to create nfqueue(IPv4, in): %s", err)
+		return fmt.Errorf("nfqueue(IPv6, out): %w", err)
 	}
-	in6Queue, err = nfqueue.NewNFQueue(17160)
+	in6Queue, err = nfQueueFactory(17160, true)
```
Raskaro, line 257: It would maybe be nice to add constants for this magic numbers ... It also works otherwise though

ppacher: :+1: for that but I didn't want risking to much with on-going changes on that part of portmaster.

PR #114, opened 2020-07-30
Fix IPv6 packet parsing

Raskaro:
Integration into the old NFQs are not done since they will be removed soon. Integration into the new NFQs are also not done yet.

### DIFF
--- a/network/packet/parse.go
+++ b/network/packet/parse.go
@@ -6,86 +6,164 @@ import (
 
 	"github.com/google/gopacket"
 	"github.com/google/gopacket/layers"
+	"github.com/safing/portbase/log"
 )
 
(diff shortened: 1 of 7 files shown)


network/packet/parse.go
```diff
@@ -6,86 +6,164 @@ import (
...
+			return err
+		}
 	}
 
-	if errLayer := parsedPacket.ErrorLayer(); errLayer != nil {
-		return errLayer.Error()
+	// 255 is reserved by IANA and used as a "failed-to-detect"
+	// marker for IPv6 (parseIPv4 always sets the protocl field)
+	if pktInfo.Protocol == 255 {
+		tryFindIPv6TransportProtocol(packet, pktInfo)
```
dhaavi, line 161: Would recommend error handling here.

PR #126, opened 2020-08-11
Add automatic upgrades for spn-hub

dhaavi:


updates/upgrader.go
```diff
@@ -128,6 +139,42 @@ func upgradeCoreNotifyActionHandler(n *notifications.Notification) {
...
+
+		// get newest spn-hub
+		new, err := GetPlatformFile(identifier)
+		if err != nil {
+			return err
+		}
+		spnHubUpdate = new
+	} else {
+		return nil
+	}
```
ppacher, line 162: Make an early return and outdent the rest of the if block
```
if spnHubUpdate != nil && !spnHubUpdate.UpgradeAvailable() {
    return nil
}
```

updates/upgrader.go
```diff
@@ -128,6 +139,42 @@ func upgradeCoreNotifyActionHandler(n *notifications.Notification) {
...
+	if info.GetInfo().Version != spnHubUpdate.Version() {
+		// get random delay with up to three hours
+		delayMinutes, err := rng.Number(3 * 60)
+		if err != nil {
+			return err
+		}
+
+		triggerRestart(time.Duration(delayMinutes) * time.Minute)
+		hubUpgradeStarted = true
+	}
```
ppacher, line 173: Maybe we should log that somewhere? If I'm watching for an hub upgrade to be downloaded it would be interesting to know when the restart is scheduled.

updates/upgrader.go
```diff
@@ -61,24 +61,35 @@ func upgrader(_ context.Context, _ interface{}) error {
 		log.Warningf("updates: failed to upgrade portmaster-start: %s", err)
 	}
 
-	if UpgradeCore {
+	binName := strings.TrimSuffix(
+		filepath.Base(os.Args[0]),
+		".exe",
+	)
+	switch binName {
+	case "portmaster-core":
```
ppacher, line 69: That check is likely to fail if the filenames are named differently. Also, at least on my machine the portmaster binary (`os.Args[0]`, `ps -ejh | grep portmaster`) is `/var/lib/portmaster/updates/linux_amd64/core/portmaster-core_v0-5-0` so the check actually fails already.

updates/restart.go
```diff
@@ -0,0 +1,48 @@
...
+func init() {
+	restartTask = module.NewTask("automatic restart", automaticRestart)
+}
+
+func triggerRestart(delay time.Duration) {
+	restartPending.Set()
+	restartTask.Schedule(time.Now().Add(delay))
+}
+
+// TriggerRestartIfPending triggers an automatic restart, if one is pending. This can be used to prepone a scheduled restart if the conditions are preferable.
```
ppacher, line 32:
```suggestion
// TriggerRestartIfPending triggers an automatic restart, if one is pending.
// This can be used to prepone a scheduled restart if the conditions are preferable.
```

updates/restart.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+func init() {
+	restartTask = module.NewTask("automatic restart", automaticRestart)
+}
+
+func triggerRestart(delay time.Duration) {
+	restartPending.Set()
+	restartTask.Schedule(time.Now().Add(delay))
+}
```
ppacher, line 30: What about exporting this so we can use in in core/events.go instead of using `updates.RestartExitCode` and `modules.Shutdown()` directly. 

```suggestion
// DelayedRestart triggers a restart of the application by shutting down the module system
// and returning with RestartExitCode. If a restart is already pending this is a no-op.
func DelayedRestart(delay time.Duration) {
    if restartPending.SetToIf(false, true) {
         restartTask.Schedule(time.Now().Add(delay))
    }
}
```

updates/upgrader.go
dhaavi: Now logged in `DelayedRestart`.

updates/restart.go
dhaavi: `DelayedRestart` is meant to be able to update an existing delay. I updated the function description.

PR #130, opened 2020-08-12
Fix and improve network listener system integration and check

dhaavi:
Also, improve logging.


netenv/addresses.go
```diff
@@ -44,3 +45,48 @@ func GetAssignedGlobalAddresses() (ipv4 []net.IP, ipv6 []net.IP, err error) {
...
+	if err != nil {
+		return false, err
+	}
+
+	// check again
+	for _, myIP := range myIPs() {
+		if ip.Equal(myIP) {
+			return true, nil
+		}
+	}
```
ppacher, line 89: Is that refresh logic really necessary? `checkMyIP` gets a func parameter (`myIPs`) to know against which set of IPs to check but assumes that it's fine to update `myIP4s` and `myIP6s`. What if the function returns something completely different? IMHO that behavior is not clear / mixed a bit. I'd rather drop the refresh part from that function (and move it to the caller if that's really a thing) or replace the `myIPs` parameter with a boolean.
AFAIK there will never be lots of IP addresses so we could actually just use one single slice that has both IPv4 and IPv6 addresses (`ip.Equal()` takes care of that anyway).

PR #133, opened 2020-08-14
Fix and improve fast tracked permits in firewall and network integration

dhaavi:


firewall/interception.go
```diff
@@ -188,6 +80,90 @@ func handlePacket(pkt packet.Packet) {
...
+			}
+
+			// Log and permit.
+			switch meta.DstPort {
+			case 53:
+				log.Debugf("accepting local dns: %s", pkt)
+			case apiPort:
+				log.Debugf("accepting api connection: %s", pkt)
+			default:
+				log.Debugf("accepting connection to internal service: %s", pkt)
```
ppacher, line 157: This once cannot happen so I'd rather drop it instead of accepting the connection

dhaavi: That is true, I viewed it more as a defensive thing in case you forget to add the port to this switch case too. But it is cleaner to just return `false` and have the normal procedure handle it.

PR #136, opened 2020-08-14
Clean up domain resolving scopes

dhaavi:
Switch to .17.home.arpa. for internal use

### DIFF
--- a/netenv/online-status.go
+++ b/netenv/online-status.go
@@ -41,7 +41,10 @@ var (
 	DNSTestDomain     = "one.one.one.one."
 	DNSTestExpectedIP = net.IPv4(1, 1, 1, 1)
 
-	SpecialCaptivePortalDomain = "captiveportal.local."
+	// SpecialCaptivePortalDomain is the domain name used to point to the detected captive portal IP
+	// or the captive portal test IP. The default value should be overridden by the resolver package,
+	// which defines the custom internal domain name to use.
+	SpecialCaptivePortalDomain = "captiveportal.invalid."
 )
 
 var (
--- a/resolver/resolver-env.go
+++ b/resolver/resolver-env.go
@@ -2,12 +2,20 @@ package resolver
 
 import (
 	"context"
+	"fmt"
 	"net"
 
 	"github.com/miekg/dns"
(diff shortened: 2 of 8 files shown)


netenv/online-status.go
```diff
@@ -41,7 +41,10 @@ var (
 	DNSTestDomain     = "one.one.one.one."
 	DNSTestExpectedIP = net.IPv4(1, 1, 1, 1)
 
-	SpecialCaptivePortalDomain = "captiveportal.local."
+	// SpecialCaptivePortalDomain is the domain name used to point to the detected captive portal IP
+	// or the captive portal test IP. The default value should be overridden by the resolver package,
+	// which defines the custom internal domain name to use.
+	SpecialCaptivePortalDomain = "captiveportal.invalid."
```
ppacher, line 47: Shouldn't that be `captiveportal.home.arpa` now?

resolver/resolver-env.go
```diff
@@ -18,67 +26,92 @@ var (
...
+		internalSpecialUseDomain,
+	))
+	if err != nil {
+		return err
+	}
+
+	internalSpecialUseComment, err = dns.NewRR(fmt.Sprintf(
+		`%s 17 IN TXT "This is a special use TLD of the Portmaster."`,
+		internalSpecialUseDomain,
+	))
```
ppacher, line 49: I thought we removed the `.17.` TLD?

resolver/scopes.go
```diff
@@ -116,99 +107,70 @@ func domainInScope(dotPrefixedFQDN string, scopeList []string) bool {
...
-		} else {
-			log.Tracef("skipping non-compliant resolver: %s", mDNSResolver.Server)
+
+	// Handle multicast domains
+	if domainInScope(q.dotPrefixedFQDN, multicastDomains) {
+		selected = addResolvers(ctx, q, selected, mDNSResolvers)
+		// Add local resolvers if no resolvers were selected.
+		if len(selected) == 0 {
+			selected = addResolvers(ctx, q, selected, localResolvers)
 		}
```
ppacher, line 139: why only add them if there are no `mDNSResolvers`. IMHO it should either always query them or not.

PR #152, opened 2020-09-21
Refactor status package to use portbase/runtime.

ppacher:
Refactor the status package to use portbase/runtime and
make system status readonly. Also adapts the code base
to the new portbase/notifications package.

This will require a rebase of #128


status/module.go
```diff
@@ -1,66 +1,36 @@
...
+
+	err := module.RegisterEventHook(
+		"netenv",
+		netenv.OnlineStatusChangedEvent,
+		"update online status in system status",
+		func(_ context.Context, _ interface{}) error {
+			triggerAutopilot()
+			return nil
+		},
+	)
```
dhaavi, line 30: Why is the autopilot triggered when the network changes?

dhaavi (approved):

status/module.go
ppacher: It won't do much here as the selected security level has not yet changed but it will push a new `SystemStatusRecord`.  Though I decided to trigger the whole autopilot since we will add "untrusted" network detection anyway.

dhaavi: The network trust system will be using Threats anyway, so that will be taken care of. But yes, we need to update for the online status.

PR #154, opened 2020-09-22
Improve DNS resolving

dhaavi:


firewall/bypassing.go
```diff
@@ -13,7 +13,9 @@ import (
 func PreventBypassing(conn *network.Connection) (endpoints.EPResult, string, nsutil.Responder) {
 	// Block firefox canary domain to disable DoH
 	if strings.ToLower(conn.Entity.Domain) == "use-application-dns.net." {
-		return endpoints.Denied, "blocked canary domain to prevent enabling DNS-over-HTTPs", nsutil.NxDomain()
+		return endpoints.Denied,
+			"blocked canary domain to prevent enabling of DNS-over-HTTPs",
+			nsutil.NxDomain("blocked canary domain to prevent enabling of DNS-over-HTTPs")
```
ppacher, line 18: I'd rather try to not duplicate the blocked message here. This could be moved to a constant as well

resolver/resolvers.go
```diff
@@ -90,6 +90,16 @@ func createResolver(resolverURL, source string) (*Resolver, bool, error) {
...
+	// Add default port for scheme if it is missing.
+	if u.Port() == "" {
+		switch u.Scheme {
+		case ServerTypeDNS, ServerTypeTCP:
+			u.Host += ":53"
+		case ServerTypeDoT:
+			u.Host += ":853"
+		}
+	}
+
```
ppacher, line 102: I guess we should update the validation regex in the config as well?

resolver/resolve.go
```diff
@@ -358,6 +358,10 @@ resolveLoop:
 				// Defensive: This should normally not happen.
 				continue
 			}
+			// Check if we got NXDomain and whether we should try another resolver.
+			if rrCache.IsNXDomain() && tryAll {
+				continue
+			}
```
ppacher, line 364: We should retry at SERVFAIL but also on NXDOMAIN? NXDOMAIN should be very clear in it's meaning. This check may also cause upstream block detection to fail (and may even cause the whole upstream block detection to be useless with `blockedIf=empty` if we always try a different resolver)

resolver/resolver-tcp.go
```diff
@@ -68,10 +72,12 @@ func NewTCPResolver(resolver *Resolver) *TCPResolver {
 			Timeout:      defaultConnectTimeout,
 			WriteTimeout: tcpWriteTimeout,
 		},
+		clientStarted:   abool.New(),
+		clientHeartbeat: make(chan struct{}),
+		clientCancel:    func() {},
 		connInstanceID:  &instanceID,
-		queries:         make(chan *dns.Msg, 100),
+		queries:         make(chan *dns.Msg, 1000),
```
ppacher, line 79: I'm not sure this is really needed....

resolver/resolver-tcp.go
```diff
@@ -168,6 +176,21 @@ func (tr *TCPResolver) Query(ctx context.Context, q *Query) (*RRCache, error) {
...
+	cancelClient := tr.clientCancel
+	tr.Unlock()
+
+	// Check if the client is alive with the heartbeat, if not shut it down.
+	select {
+	case tr.clientHeartbeat <- struct{}{}:
+	case <-time.After(defaultRequestTimeout):
+		cancelClient()
+	}
+}
```
ppacher, line 192: we should at least log a message in the timeout case here

ppacher (changes requested):

resolver/resolvers.go
dhaavi: The validation is very generous anyway and already includes this change. Unless we want to improve it?

resolver/resolve.go
dhaavi: The important tidbit is the `tryAll` which is only true in special cases. See `scopes.go`.

resolver/resolvers.go
ppacher: You're right the validation regex is already lax enough. Though, we should update the help text: https://github.com/safing/portmaster/blob/develop/resolver/config.go#L99

firewall/bypassing.go
```diff
@@ -15,7 +15,7 @@ func PreventBypassing(conn *network.Connection) (endpoints.EPResult, string, nsu
 	if strings.ToLower(conn.Entity.Domain) == "use-application-dns.net." {
 		return endpoints.Denied,
 			"blocked canary domain to prevent enabling of DNS-over-HTTPs",
-			nsutil.NxDomain("blocked canary domain to prevent enabling of DNS-over-HTTPs")
+			nsutil.NxDomain("")
```
ppacher: Was that intended? I though about something like:
```go
blockReason = "blocked canary domain to prevent ..."
return entpoins.Denied, blockReason, nsutil.NxDomain(blockReason)
```

ppacher (approved):

firewall/bypassing.go
dhaavi: Yes, the reason is now always added to the response messages using the interface of the Connection itself. The argument to the responders is only for special cases within the nameserver.

ppacher: Maybe we should change it to `NxDomain(...string)` then so you don't need to pass in an empty string all the time?

resolver/resolve.go
```diff
@@ -195,27 +201,36 @@ func checkCache(ctx context.Context, q *Query) *RRCache {
...
 		module.StartWorker("resolve async", func(ctx context.Context) error {
-			_, err := resolveAndCache(ctx, q)
+			ctx, tracer := log.AddTracer(ctx)
+			tracer.Debugf("resolver: resolving %s async", q.ID())
+			_, err := resolveAndCache(ctx, q, nil)
 			if err != nil {
-				log.Warningf("resolver: async query for %s%s failed: %s", q.FQDN, q.QType, err)
+				tracer.Warningf("resolver: async query for %s failed: %s", q.ID(), err)
 			}
 			return nil
```
ppacher, line 235: We never log the success of the async resolving. I'd add a Info Debug level message to the tracer containg the result. This way we can better spot timing issues with i.e. upstream resolvers.

ppacher (approved):

nameserver/nsutil/nsutil.go
```diff
@@ -53,40 +58,134 @@ func ZeroIP() ResponderFunc {
...
 // Refused returns a ResponderFunc that replies with REFUSED.
-func Refused() ResponderFunc {
-	return func(query *dns.Msg, _ string, _ interface{}) *dns.Msg {
-		return new(dns.Msg).SetRcode(query, dns.RcodeRefused)
+func Refused(msgs ...string) ResponderFunc {
+	return func(ctx context.Context, request *dns.Msg) *dns.Msg {
+		reply := new(dns.Msg).SetRcode(request, dns.RcodeRefused)
+		for _, msg := range msgs {
+			AddMessageToReply(ctx, reply, log.InfoLevel, msg)
+		}
```
ppacher, line 144: That's repeated over and over again. Maybe make `AddMessagesToReply` accept `...string` itself?

resolver/resolve.go
```diff
@@ -392,21 +392,32 @@ resolveLoop:
...
+		switch {
+		case err != nil:
+			// There was an error during resolving, return the old cache entry instead.
+			log.Tracer(ctx).Debugf("resolver: serving backup cache of %s because query failed: %s", q.ID(), err)
+			return oldCache, nil
+		case rrCache.IsNXDomain():
+			// The new result is NXDomain, return the old cache entry instead.
+			log.Tracer(ctx).Debugf("resolver: serving backup cache of %s because fresh response is NXDomain", q.ID())
+			return oldCache, nil
+		}
```
ppacher: I know this shouldn't harm much but we should returns NXDOMAIN to the user instead of serving a stale record as the IP address is likely not correct anymore. We should serve the cache for SERVFAIL but IMHO not for NXDOMAIN.

ppacher (changes requested):

ppacher (approved):

ppacher (dismissed):

ppacher (approved):

PR #162, opened 2020-09-29
templates/bug-report.md: fixed windows logs path

northys:


dhaavi (approved):
Thanks for the fix! 👍

PR #163, opened 2020-09-29
Support non-standard format on dns query name

dhaavi:


nameserver/nameserver.go
```diff
@@ -243,6 +252,11 @@ func handleRequest(ctx context.Context, w dns.ResponseWriter, request *dns.Msg)
 	// Save dns request as open.
 	defer network.SaveOpenDNSRequest(conn)
 
+	// Revert back to non-standard question format, if we had to convert.
+	if nonStandardQuestionFormat {
+		rrCache.ReplaceAnswerNames(originalQuestion.Name)
+	}
+
```
ppacher, line 259: Shouldn't we clone the rrCache here before replacing the answers inline? The rrCache can be hold by the caching layer of the database and thus would interfere with others that used different folding in the query name. Also, `ReplaceAnserNames` does not lock the rrCache which seems to be racy with other queries served from cache.

ppacher (changes requested):

ppacher (approved):

PR #167, opened 2020-10-01
Add simple packet metrics

ppacher:
Adds a simple `--packet-metrics <file-path>` parameter to dump packet metrics every few seconds as a CSV file

### DIFF
--- a/firewall/interception/interception.go
+++ b/firewall/interception/interception.go
@@ -25,18 +25,18 @@ func Start() error {
 		return nil
 	}
 
-	var ch = Packets
+	var inputPackets = Packets
 	if packetMetricsDestination != "" {
 		go metrics.writeMetrics()
-		ch = make(chan packet.Packet)
+		inputPackets = make(chan packet.Packet)
 		go func() {
-			for p := range ch {
+			for p := range inputPackets {
 				Packets <- tracePacket(p)
 			}
 		}()
 	}
 
-	return start(ch)
+	return start(inputPackets)
 }
 
 // Stop starts the interception.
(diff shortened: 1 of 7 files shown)


firewall/interception/introspection.go
```diff
@@ -0,0 +1,78 @@
...
+
+var (
+	packetMetricsDestination string
+	metrics                  = &packetMetrics{
+		done: make(chan struct{}),
+	}
+)
+
+func init() {
+	flag.StringVar(&packetMetricsDestination, "packet-metrics", "", "Write packet metrics")
```
dhaavi, line 21: Would suggest naming the flag `write-packet-metrics` and changing the description to `Write packet metrics to the specified file`. This makes it clear that these are written to a file.

firewall/interception/introspection.go
```diff
@@ -0,0 +1,78 @@
...
+	"os"
+	"sync"
+	"time"
+
+	"github.com/safing/portbase/log"
+)
+
+var (
+	packetMetricsDestination string
+	metrics                  = &packetMetrics{
```
dhaavi, line 15: For some reason the CI failed, but this would have been definitely flagged. ;)

firewall/interception/interception.go
```diff
@@ -25,7 +25,18 @@ func Start() error {
 		return nil
 	}
 
-	return start()
+	var ch = Packets
```
dhaavi, line 28: Would suggest to use a more descriptive name for `ch` everywhere, such as `packetInput` or `packetIngest`.

dhaavi (changes requested):
Very nice and simple way to collect packet timing data! 👌 Some minor comments.

dhaavi (approved):

PR #170, opened 2020-10-07
Fixes race conditions due to incorrect usage of PushUpdate or PushFunc

ppacher:


network/connection.go
```diff
@@ -19,48 +18,114 @@ import (
...
-	pktQueue        chan packet.Packet
+	// Inspecting is set to true if the connection is being inspected
+	// by one or more of the registered inspectors. This property may
+	// be changed during the lifetime of a connection and must be guarded
+	// using the connection lock.
+	Inspecting bool
+	// Tunneled is currently unused and MUST be ignored.
+	Tunneled bool
+	// Encrypted is currently unused and MUST be ignored.
+	Encrypted bool
```
dhaavi, line 100: Both of these are currently being worked on. Should we not add a description already?

dhaavi (approved):
Very nice!

PR #174, opened 2020-10-13
Add scoping to IPInfo

dhaavi:


firewall/dns.go
```diff
@@ -205,30 +205,49 @@ func mayBlockCNAMEs(conn *network.Connection) bool {
 	return false
 }
 
+// updateIPsAndCNAMEs saves all the IP->Name mappings to the cache database and
+// updates the CNAMEs in the Connection's Entity.
 func updateIPsAndCNAMEs(q *resolver.Query, rrCache *resolver.RRCache, conn *network.Connection) {
-	// save IP addresses to IPInfo
+	// FIXME: ignore localhost
+
```
ppacher, line 212: I guess this should be done?

firewall/dns.go
```diff
@@ -205,30 +205,49 @@ func mayBlockCNAMEs(conn *network.Connection) bool {
...
+	// Get IPInfo scope.
+	var scope string
+	proc := conn.Process()
+	if proc != nil {
+		scope = proc.LocalProfileKey
+	}
+
```
ppacher, line 219: I'm confused about our scopes. I thought that can either be a domain or one of those two-letter constants? Now it can be a profile key as well?

I guess this scope is something else so maybe rename the variable?

resolver/ipinfo.go
```diff
@@ -54,65 +72,61 @@ func (rds ResolvedDomains) String() string {
...
 	IP string
 
-	// Domains holds a list of domains that have been
-	// resolved to IP. This field is deprecated and should
-	// be removed.
-	// DEPRECATED: remove with alpha.
-	Domains []string `json:"Domains,omitempty"`
+	// Scope holds a scope for this IPInfo.
+	// Usually this would be the Profile ID of the associated process.
+	Scope string
```
ppacher, line 85: Please use a different name here! The `Scope` in the connection is already confusing enough. Just call this one `ProfileID`?

resolver/ipinfo.go
```diff
@@ -150,27 +152,38 @@ func GetIPInfo(ip string) (*IPInfo, error) {
...
 	}
+	info.UpdateMeta()
+	expires += 3600 // Add one hour to expiry as a buffer.
+	info.Meta().SetAbsoluteExpiry(expires)
+
+	info.Unlock()
 
-	return ipInfoDatabase.Put(ipi)
+	return ipInfoDatabase.Put(info)
 }
```
ppacher, line 181: `info` should be locked when calling `Put` so defer the unlock at the beginning

resolver/namerecord.go
```diff
@@ -32,7 +46,8 @@ type NameRecord struct {
 	Answer   []string
 	Ns       []string
 	Extra    []string
-	TTL      int64
+	// TODO: Name change in progress. Rename "TTL" field to "Expires" in Q1 2021.
+	TTL int64 `json:"Expires"`
```
ppacher, line 51: If we ever start writting queries for the NameRecord than https://github.com/safing/portbase/issues/75 will bite us! Why not rename the field right now? It's a cache database anyway...

ppacher (changes requested):

resolver/ipinfo.go
dhaavi: Ah, no. `Put` actually does lock the record.

ppacher (approved):

PR #175, opened 2020-10-13
Fix nfq-exp flag and nfq logging messages

dhaavi:


firewall/interception/nfqueue_linux.go
```diff
@@ -36,10 +34,10 @@ var (
 )
 
 func init() {
-	flag.BoolVar(&experimentalNfqueueBackend, "experimental-nfqueue", true, "(deprecated flag; always used)")
+	flag.BoolVar(&experimentalNfqueueBackend, "experimental-nfqueue", false, "(deprecated flag; always used)")
```
ppacher, line 37: Why did you change that to false? It's always used that's what it defaults to `true`? (although, it doesn't matter because the flag isn't used anyway)

ppacher (approved):

firewall/interception/nfqueue_linux.go
dhaavi: Because it then always shows the deprecated warnings.

PR #176, opened 2020-10-13
Fix domain endpoint validation

dhaavi:


profile/endpoints/endpoint-domain.go
```diff
@@ -88,53 +88,72 @@ func (ep *EndpointDomain) String() string {
...
+	if containsRegex.MatchString(value) {
+		return true
+	}
+
+	// Remove special leading characters.
+	value = strings.TrimLeft(value, ".*")
+
+	// Check if value is now a valid domain.
+	return netutils.IsValidFqdn(value)
+}
```
ppacher, line 107: This will fail if the domain can still have a `.*` suffix. The suffix is trimmed a few lines further down below after the call to `isValidEndpointDomain`

Also, the comments about a "contains" value and the "special leading characters" are a bit confusing

profile/endpoints/endpoint-domain.go
```diff
@@ -88,53 +88,72 @@ func (ep *EndpointDomain) String() string {
...
-		default:
-			domain += "."
-		}
+	// fix domain ending
+	switch domain[len(domain)-1] {
+	case '.':
+	case '*':
+	default:
+		domain += "."
+	}
```
ppacher, line 122:
```suggestion
	switch domain[len(domain)-1] {
	case '.', '*':
	default:
		domain += "."
	}
```

profile/endpoints/endpoint-domain.go
```diff
@@ -88,53 +88,72 @@ func (ep *EndpointDomain) String() string {
...
+	case strings.HasPrefix(domain, "*"):
+		ep.MatchType = domainMatchTypeSuffix
+		ep.Domain = strings.Trim(domain, "*")
+		return ep.parsePPP(ep, fields)
+
+	case strings.HasPrefix(domain, "."):
+		ep.MatchType = domainMatchTypeZone
+		ep.Domain = strings.TrimLeft(domain, ".")
+		ep.DomainZone = "." + ep.Domain
+		return ep.parsePPP(ep, fields)
```
ppacher, line 152: I'd suggest to use the correct trim function here depending on Prefix/Suffix. It currently works because it's ensured that there's only one prefix or suffix but it looks like a bug.

ppacher (changes requested):

profile/endpoints/endpoint-domain.go
```diff
@@ -88,72 +88,65 @@ func (ep *EndpointDomain) String() string {
...
+	// Validate domain "content".
+	switch {
+	case needValidFQDN && !netutils.IsValidFqdn(ep.Domain):
+		return nil, nil
+	case !needValidFQDN && !allowedDomainChars.MatchString(ep.Domain):
+		return nil, nil
+	case strings.Contains(ep.Domain, ".."):
+		// The above regex does not catch double dots.
+		return nil, nil
+	}
```
ppacher: I'd recomment to not use switches for complex boolean cases because it's really hard to reason about them and using nested if-else blocks is way more easy to read.

IMHO using naked switches is awesome for single-conditions but as soon as one starts combining conditions with `&&` or `||` reading and understanding a switch can be hard. 

```suggestion
	if needValidFQDN {
		if !netutils.IsValidFqdn(ep.Domain) {
			return nil, nil
		}
	} else {
		if !allowedDomainChars.MatchString(ep.Domain) {
			return nil, nil
		}
	}
	
	if strings.Contains(ep.Domain, "..") {
		return nil, nil
	}
```

ppacher (approved):

PR #180, opened 2020-10-16
Improve concurrent connection handling support

dhaavi:


network/proc/findpid.go
```diff
@@ -5,99 +5,100 @@ package proc
...
 func GetPID(socketInfo socket.Info) (pid int) {
-	pidsByUserLock.Lock()
-	defer pidsByUserLock.Unlock()
-
-	if socketInfo.GetPID() != socket.UnidentifiedProcessID {
-		return socketInfo.GetPID()
+	// Get currently assigned PID to the socket info.
+	socketInfoLock.RLock()
+	currentPid := socketInfo.GetPID()
+	socketInfoLock.RUnlock()
```
ppacher, line 29: Why do you use a global mutex for accessing `GetPID()` from a struct thats defined in a completely different package? sockeInfo needs to holds the struct itself otherwise we will always have race-conditions because nobody outside of the `proc` package can use this lock

network/proc/findpid.go
```diff
@@ -5,99 +5,100 @@ package proc
...
-			for _, possiblePID := range pids {
-				// only check if not already checked
-				if sort.SearchInts(checkedUserPids, possiblePID) == len {
-					if findSocketFromPid(possiblePID, inode) {
-						return possiblePID
+			pids, ok = getPidsByUser(uid)
+			if ok {
+				for _, pid = range pids {
+					if findSocketFromPid(pid, socketName) {
+						return pid
```
ppacher, line 81: This is the exact same code block as above and we would need to keep them in sync. What about doing an `continue`? That would be much more clear to me.

network/proc/findpid.go
```diff
@@ -5,99 +5,100 @@ package proc
...
-	for possibleUID, pids := range pidsByUser {
-		if possibleUID != uid {
-			for _, possiblePID := range pids {
-				if findSocketFromPid(possiblePID, inode) {
-					return possiblePID
-				}
-			}
+		// We have updated the PID map, but still cannot find anything.
+		// So, there is nothing we can other than wait a little for the kernel to
+		// populate the information.
```
ppacher, line 89: There's a word missing

network/proc/pids_by_user.go
```diff
@@ -0,0 +1,82 @@
...
+	"strconv"
+	"sync"
+	"syscall"
+
+	"github.com/safing/portbase/log"
+	"github.com/safing/portbase/utils"
+)
+
+var (
+	// pidsByUserLock is also used for locking the socketInfo.PID on all socket.*Info structs.
```
ppacher, line 15: As mentioned in my other comment, this is really really error prone as we are trying to protect a struct/interface that is not in the domain of the proc package.

network/proc/pids_by_user.go
```diff
@@ -0,0 +1,82 @@
...
+			}
+			pidCnt++
+		}
+
+		// Reverse slice orders, because higher PIDs will be more likely to be searched for.
+		for _, slice := range newPidsByUser {
+			for i, j := 0, len(slice)-1; i < j; i, j = i+1, j-1 {
+				slice[i], slice[j] = slice[j], slice[i]
+			}
+		}
```
ppacher, line 74: instead of reversing the slice we could also search bottom-to-top which would safe us all this interations. Since we are "re-loading" processes if we don't find the PID immediately we actually do this "reversing" quite often.

ppacher (changes requested):

network/proc/findpid.go
```diff
@@ -63,9 +57,11 @@ func findPID(uid, inode int) (pid int) {
...
 		// If we have found PIDs, search them.
 		if ok {
-			for _, pid = range pids {
-				if findSocketFromPid(pid, socketName) {
-					return pid
+			// Look through the PIDs in reverse order, because higher/newer PIDs will be more likely to
+			// be searched for.
+			for i := len(pids) - 1; i > 0; i-- {
+				if findSocketFromPid(pids[i], socketName) {
+					return pids[i]
```
ppacher, line 64:
```suggestion
			for i := len(pids) - 1; i >= 0; i-- {
				if findSocketFromPid(pids[i], socketName) {
					return pids[i]
```
Same issue a few lines below

network/proc/findpid.go
```diff
@@ -24,22 +21,19 @@ var (
...
-	socketInfoLock.RUnlock()
 
 	// If the current PID already is valid (ie. not unidentified), return it immediately.
 	if currentPid != socket.UnidentifiedProcessID {
 		return currentPid
 	}
 
 	// Find PID for the given UID and inode.
-	pid = findPID(socketInfo.GetUID(), socketInfo.GetInode())
+	// uid, inode := socketInfo.GetUIDandInode()
```
ppacher, line 32: I guess this line can be removed

ppacher (approved):

ppacher:
One small issue, otherwise LGTM

ppacher (approved):

PR #183, opened 2020-10-29
Revamp profile and process handling, add optionKey to reason

dhaavi:


process/database.go
```diff
@@ -106,32 +106,33 @@ func CleanProcessStorage(activePIDs map[int]struct{}) {
...
-			// internal
-		case p.Pid == SystemProcessID:
-			// internal
-		case active:
-			// process in system process table or recently seen on the network
+		case p.LastSeen == 0:
+			// add last
+			p.LastSeen = time.Now().Unix()
+		case p.LastSeen > threshold:
+			// within keep period
```
ppacher, line 130: Event though the PID is immutable we are mutating LastSeen so I guess the profile should be locked?

profile/get.go
```diff
@@ -0,0 +1,202 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		// Process profiles coming directly from the database.
+		// As we don't use any caching, these will be new objects.
+
+		// Mark the profile as being saved internally in order to bypass checks.
+		profile.internalSave = true
+
```
ppacher, line 102: Why is that required?

profile/get.go
```diff
@@ -0,0 +1,202 @@
...
+		),
+	)
+	if err != nil {
+		return nil, false, err
+	}
+
+	// Only wait for the first result, or until the query ends.
+	r := <-it.Next
+	// Then cancel the query, should it still be running.
+	it.Cancel()
```
ppacher, line 160: Maybe better defer that so that it will always be canceled even in case of a panic?

profile/get.go
```diff
@@ -0,0 +1,202 @@
...
+	// If there was no profile in the database, create a new one, and return it.
+	profile = New(SourceLocal, "")
+	profile.LinkedPath = linkedPath
+
+	// Check if the profile should be marked as internal.
+	// This is the case whenever the binary resides within the data root dir.
+	if strings.HasPrefix(linkedPath, dataroot.Root().Dir+string(os.PathSeparator)) {
+		profile.Internal = true
+	}
+
```
ppacher, line 177: On windows and some linux file-system types the paths are not case-sensitive so this check may not always work. At least for Windows we should consider checking with equal fold. For linux this is a bit dangerous ...

profile/profile-layered.go
```diff
@@ -56,7 +62,8 @@ func NewLayeredProfile(localProfile *Profile) *LayeredProfile {
 	new := &LayeredProfile{
 		localProfile:       localProfile,
 		layers:             make([]*Profile, 0, len(localProfile.LinkedProfiles)+1),
-		revisionCounter:    0,
+		LayerIDs:           make([]string, 0, len(localProfile.LinkedProfiles)+1),
+		RevisionCounter:    0,
```
ppacher, line 66: `RevisionCounter` is set to it's zero value. I'd rather remove it

profile/profile-layered.go
```diff
@@ -308,6 +360,33 @@ func (lp *LayeredProfile) wrapSecurityLevelOption(configKey string, globalConfig
...
+				}
+			}
+			if !found {
+				value = globalConfig()
+			}
+		}
+
+		return value
+	}
+}
```
ppacher, line 388: As discussed, this is racy

ppacher (changes requested):

profile/get.go
dhaavi: Discussed, and it's going stay for this right now, as only false-positives might be problem for UX. There is no firewall related change anyway.

profile/get.go
dhaavi: Updated comment.

profile/get.go
dhaavi: It would only move one line up and I think it's better to kill asap, as the function also does validation, which takes longer. We can save database resources by cancelling asap. If `<-it.Next` panics, `it.Cancel()` might just as well.

ppacher (approved):

ppacher (approved):

PR #195, opened 2020-11-17
Adapt profiles to use new binary metadata system

dhaavi:


profile/module_test.go
```diff
@@ -1,11 +1,7 @@
...
-	"testing"
-
-	"github.com/safing/portmaster/core/pmtesting"
-)
-
+/*
 func TestMain(m *testing.M) {
 	pmtesting.TestMain(m, module)
 }
+*/
```
ppacher, line 7: Maybe we should remove the whole file instead so we are not "tricked" by "everything's fine because there's a test-file".

profile/profile.go
```diff
@@ -1,13 +1,17 @@
...
 	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
 
+	"github.com/safing/portbase/utils/osdetail"
+
 	"github.com/tevino/abool"
 
 	"github.com/safing/portbase/config"
```
ppacher, line 17: Looks like `gofmt` did not reorder the imports. the `safing/portbase` imports belong together

profile/profile.go
```diff
@@ -377,3 +382,106 @@ func EnsureProfile(r record.Record) (*Profile, error) {
...
+	// Update profile name if it is empty or equals the filename, which is the
+	// case for older profiles.
+	if p.Name == "" || p.Name == filename {
+		// Generate a default profile name if does not exist.
+		p.Name = osdetail.GenerateBinaryNameFromPath(p.LinkedPath)
+		if p.Name == filename {
+			// TODO: Theoretically, the generated name could be identical to the
+			// filename.
+			// As a quick fix, append a space to the name.
+			p.Name += " "
```
ppacher, line 425: Why do we need this "quick fix"

ppacher (approved):

PR #206, opened 2020-11-30
Shorten caching for empty success responses.

dhaavi:
Note:
```
if !rrCache.Cacheable() {
	return
}
```
is included in `rrCache.Save()`

### DIFF
--- a/resolver/rrcache.go
+++ b/resolver/rrcache.go
@@ -85,6 +85,12 @@ func (rrCache *RRCache) Clean(minExpires uint32) {
 		lowestTTL = maxTTL
 	}
 
+	// Adjust return code if there are no answers
+	if rrCache.RCode == dns.RcodeSuccess &&
+		len(rrCache.Answer) == 0 {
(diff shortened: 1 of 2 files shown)


resolver/rrcache.go
```diff
@@ -85,6 +85,12 @@ func (rrCache *RRCache) Clean(minExpires uint32) {
 		lowestTTL = maxTTL
 	}
 
+	// Adjust return code if there are no answers
+	if rrCache.RCode == dns.RcodeSuccess &&
+		len(rrCache.Answer) == 0 {
+		rrCache.RCode = dns.RcodeNameError
+	}
+
```
ppacher, line 93: Should we document when this can actually happen? Because AFAIK we return an error if the response gets empty because we removed all out-of-scope answers. Is there any other situation we can cause this (except from an invalid upstream-answer)?

ppacher (commented):
Otherwise LGTM

PR #227, opened 2021-01-07
Fix filterlist index updating

dhaavi:


ppacher:
I get that we missed to check for an update in `performUpdate` but why is the change to `updateListIndex` required?

dhaavi:
It's not required for it to work, but it would load the file and save the index to the database again _every_ update check.

ppacher (approved):

PR #233, opened 2021-01-19
Add support for network service

dhaavi:


ppacher (approved):
Awesome work! Looking forward to test that :)

PR #238, opened 2021-01-25
Simplify profile reloading

dhaavi:
Also, increase prompt decision timeout.

### DIFF
--- a/profile/active.go
+++ b/profile/active.go
@@ -58,6 +58,13 @@ func addActiveProfile(profile *Profile) {
 	activeProfilesLock.Lock()
 	defer activeProfilesLock.Unlock()
 
+	// Mark any previous profile as outdated.
+	previous, ok := activeProfiles[profile.ScopedID()]
+	if ok {
+		previous.outdated.Set()
+	}
+
+	// Mark new profile active and add to active profiles.
 	profile.MarkStillActive()
 	activeProfiles[profile.ScopedID()] = profile
 }
--- a/profile/get.go
+++ b/profile/get.go
@@ -28,7 +28,7 @@ const (
 
 var getProfileSingleInflight singleflight.Group
 
-// GetProfile fetches a profile. This function ensure that the profile loaded
+// GetProfile fetches a profile. This function ensures that the loaded profile
 // is shared among all callers. You must always supply both the scopedID and
 // linkedPath parameters whenever available.
 func GetProfile(source profileSource, id, linkedPath string) ( //nolint:gocognit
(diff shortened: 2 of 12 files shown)


profile/active.go
```diff
@@ -58,6 +58,13 @@ func addActiveProfile(profile *Profile) {
 	activeProfilesLock.Lock()
 	defer activeProfilesLock.Unlock()
 
+	// Mark any previous profile as outdated.
+	previous, ok := activeProfiles[profile.ScopedID()]
+	if ok {
+		previous.outdated.Set()
+	}
```
ppacher, line 65: `previous` and `ok` are not required in the outer scope so consider:
```suggestion
if previous, ok := activeProfiles[profile.ScopedID()]; ok {
    previous.outdated.Set()
}
```

profile/get.go
```diff
@@ -28,7 +28,7 @@ const (
 
 var getProfileSingleInflight singleflight.Group
 
-// GetProfile fetches a profile. This function ensure that the profile loaded
+// GetProfile fetches a profile. This function ensures that the loaded profile
 // is shared among all callers. You must always supply both the scopedID and
 // linkedPath parameters whenever available.
 func GetProfile(source profileSource, id, linkedPath string) ( //nolint:gocognit
```
ppacher, line 34: IMHO the function documentation does not explain what happes if I don't have the `scopeID` or `linkedPath available`.
We may also rename the type of the first parameter and remove the `profile` prefix as this is the package name and stutters when (trying) to use from somewhere else: `profile.profileSource`

profile/get.go
```diff
@@ -266,13 +192,13 @@ func prepProfile(r record.Record) (*Profile, error) {
...
-		log.Warningf("profiles: profile %s has (partly) invalid configuration: %s", profile.ID, err)
+		log.Errorf("profiles: profile %s has (partly) invalid configuration: %s", profile.ID, err)
 	}
 
 	// parse config
 	err = profile.parseConfig()
 	if err != nil {
-		log.Warningf("profiles: profile %s has (partly) invalid configuration: %s", profile.ID, err)
+		log.Errorf("profiles: profile %s has (partly) invalid configuration: %s", profile.ID, err)
 	}
```
ppacher, line 202: If that is an error I guess we should `return nil, err`?

profile/profile.go
```diff
@@ -301,18 +298,6 @@ func (profile *Profile) addEndpointyEntry(cfgKey, newEntry string) {
...
-		if !changed || profile.layeredProfile == nil {
-			return
-		}
-
-		profile.layeredProfile.Lock()
-		defer profile.layeredProfile.Unlock()
-
-		profile.layeredProfile.RevisionCounter++
-	}()
-
```
ppacher, line 315: How does the revision counter get handled now?

dhaavi: Profiles are marked as outdated when they are saved. The next operation that uses the profile checks for outdated profiles, reloads them and increases the counter. Only having a single way of doing that makes this simpler and less error prone.

profile/get.go
dhaavi: I improved the documentation.
Yes, `profile.profileSource` stutters a bit, but just using `source` is not explicit enough, IMHO - I tried it.

profile/get.go
dhaavi: Let's talk about that.

PR #248, opened 2021-02-05
Fix failing build by updating firewall package to new function name

ppacher:


dhaavi (approved):
Thanks, I fixed this locally during staging, but haven't had the time to do a PR yet.

PR #251, opened 2021-02-11
Improve internal profile handling

dhaavi:


network/connection.go
```diff
@@ -248,7 +254,14 @@ func NewConnectionFromExternalDNSRequest(ctx context.Context, fqdn string, cname
...
 		Started:        timestamp,
 		Ended:          timestamp,
-	}, nil
+	}
+
+	// Inherit internal status of profile.
+	if localProfile := remoteHost.Profile().LocalProfile(); localProfile != nil {
+		dnsConn.Internal = localProfile.Internal
+	}
+
```
ppacher, line 263: Can a host in network-mode even be considered "internal"? I understand that we need to inherit the flag but is this really something that can happen?

network/connection.go
```diff
@@ -352,6 +366,13 @@ func NewConnectionFromFirstPacket(pkt packet.Packet) *Connection {
 		Started:                time.Now().Unix(),
 		ProfileRevisionCounter: proc.Profile().RevisionCnt(),
 	}
+
+	// Inherit internal status of profile.
+	if localProfile := proc.Profile().LocalProfile(); localProfile != nil {
+		newConn.Internal = localProfile.Internal
+	}
+
```
ppacher, line 374: Now that I see it the third time I start thinking if we should create a dedicated factory function/helper for those inheritance parts. Something like `newConn.inherit(profile)`. Or maybe a helper `NewInhertitedConnection(profile Profile) *Connection`

process/module.go
```diff
@@ -1,17 +1,30 @@
...
 
 func prep() error {
 	return registerConfiguration()
 }
+
+func start() error {
+	updatesPath = updates.RootPath() + string(os.PathSeparator)
+	if updatesPath != "" {
+		updatesPath += string(os.PathSeparator)
+	}
```
ppacher, line 27: that cannot happen as updatePath is either a full path or "/" but never empty

process/profile.go
```diff
@@ -38,6 +39,16 @@ func (p *Process) GetProfile(ctx context.Context) (changed bool, err error) {
...
 		profileID = profile.PortmasterProfileID
+	default:
+		// Check if this is another Portmaster component.
+		if updatesPath != "" && strings.HasPrefix(p.Path, updatesPath) {
+			switch {
+			case strings.Contains(p.Path, "portmaster-app"):
+				profileID = profile.PortmasterAppProfileID
+			case strings.Contains(p.Path, "portmaster-notifier"):
+				profileID = profile.PortmasterNotifierProfileID
+			}
```
ppacher, line 50: what about a `default`. Do we expect any other valid-but-ignored usecases? Should we log/mention them here?

profile/module.go
```diff
@@ -1,17 +1,21 @@
...
+
 	"github.com/safing/portbase/log"
 
 	"github.com/safing/portbase/modules"
 
 	// module dependencies
 	_ "github.com/safing/portmaster/core/base"
+	"github.com/safing/portmaster/updates"
 	_ "github.com/safing/portmaster/updates" // dependency of semi-dependency filterlists
 )
```
ppacher, line 14: Github Actions already reported it, `portmaster/updates` is imported twice

profile/module.go
```diff
@@ -33,6 +37,11 @@ func prep() error {
 }
 
 func start() error {
+	updatesPath = updates.RootPath() + string(os.PathSeparator)
+	if updatesPath != "" {
+		updatesPath += string(os.PathSeparator)
+	}
+
```
ppacher, line 44: Same issue as in process/module.go

process/profile.go
```diff
@@ -38,6 +39,16 @@ func (p *Process) GetProfile(ctx context.Context) (changed bool, err error) {
 		profileID = profile.SystemProfileID
 	case ownPID:
 		profileID = profile.PortmasterProfileID
+	default:
+		// Check if this is another Portmaster component.
+		if updatesPath != "" && strings.HasPrefix(p.Path, updatesPath) {
+			switch {
```
ppacher, line 45: Maybe consider using `filepath.ToSlash()` to ensure the comparision uses the same separator style (`/` is also a valid separator on windows).
Also, the `updatesPath != ""` should be `updatesPath != "/"` as that's the "empty" value. See the issue in process/module.go

I.e. 
```suggestion
		// Check if this is another Portmaster component.
		if strings.HasPrefix(filepath.ToSlash(p.Path), filepath.ToSlash(updatesPath)) {
			switch {
```

profile/get.go
```diff
@@ -177,7 +156,7 @@ func findProfile(linkedPath string) (profile *Profile, err error) {
 
 	// Check if the profile should be marked as internal.
 	// This is the case whenever the binary resides within the data root dir.
-	if strings.HasPrefix(linkedPath, dataroot.Root().Dir+string(os.PathSeparator)) {
+	if updatesPath != "" && strings.HasPrefix(linkedPath, updatesPath) {
 		profile.Internal = true
```
ppacher, line 160: Same issue as above in process/profile.go

profile/profile.go
```diff
@@ -394,7 +394,7 @@ func (profile *Profile) UpdateMetadata(processName, binaryPath string) (changed
 	}
 
 	// Update LinkedPath if if differs from the process path.
-	// This will (at the moment) only be the case for the Portmaster profile.
+	// This will be the case for profiles that are assigned in a special way.
 	if profile.LinkedPath != binaryPath {
 		profile.LinkedPath = binaryPath
```
ppacher, line 399: Hm, if I ever hit this part while hunting a bug it would be nice to get reminded of those "special ways"

profile/special.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+func getSpecialProfile(profileID, linkedPath string) *Profile {
+	switch profileID {
+	case UnidentifiedProfileID:
+		return New(SourceLocal, UnidentifiedProfileID, linkedPath)
+
+	case SystemProfileID:
+		return New(SourceLocal, SystemProfileID, linkedPath)
+
```
ppacher, line 27: For consistency with the other cases, shouldn't we set a `profile.Name` here as well?

profile/special.go
```diff
@@ -0,0 +1,61 @@
...
+		profile := New(SourceLocal, PortmasterNotifierProfileID, linkedPath)
+		profile.Name = "Portmaster Notifier"
+		profile.Internal = true
+		profile.Config = map[string]interface{}{
+			CfgOptionDefaultActionKey: "block",
+			CfgOptionEndpointsKey: []string{
+				"+ Localhost",
+			},
+		}
+		return profile
```
ppacher, line 56: Nice!

ppacher (changes requested):

PR #252, opened 2021-02-11
Stop randomizing dns answer records

dhaavi:
As it seems clients are even dumber when parsing the answers. There _may_ be an RFC for that, which I am not aware of, though.


ppacher:
LGTM

PR #263, opened 2021-02-23
Fix bulk profile updates via prompts

dhaavi:


firewall/prompt.go
```diff
@@ -125,9 +128,35 @@ func createPrompt(ctx context.Context, conn *network.Connection, pkt packet.Pack
 
 	// If there already is a notification, just update the expiry.
 	if n != nil {
-		n.Update(expires)
-		log.Tracer(ctx).Debugf("filter: updated existing prompt notification")
-		return
+		// Get notification state
+		n.Lock()
+		state := n.State
+		n.Unlock()
```
ppacher, line 134: copy n.SelectionActionID here as well so we can avoid re-locking the notification again below.

firewall/prompt.go
```diff
@@ -125,9 +128,35 @@ func createPrompt(ctx context.Context, conn *network.Connection, pkt packet.Pack
...
+		if action != "" {
+			switch action {
+			case allowDomainAll, allowDomainDistinct, allowIP, allowServingIP:
+				conn.Accept("permitted via prompt", profile.CfgOptionEndpointsKey)
+			default: // deny
+				conn.Deny("blocked via prompt", profile.CfgOptionEndpointsKey)
+			}
+			return nil // Do not take further action.
+		}
```
ppacher, line 156: We already handle the response around line 61 where we are waiting on `n.Response()` channel. Why do we need to check if there's an action set here as well?

ppacher (changes requested):

ppacher (approved):
LGTM

PR #272, opened 2021-03-10
Minor fixes and improvements

dhaavi:
Requires safing/portbase#120.


process/config.go
```diff
@@ -17,7 +17,7 @@ func registerConfiguration() error {
 	err := config.Register(&config.Option{
 		Name:           "Process Detection",
 		Key:            CfgOptionEnableProcessDetectionKey,
-		Description:    "This option enables the attribution of network traffic to processes. This should always be enabled, and effectively disables app profiles if disabled.",
+		Description:    "This option enables the attribution of network traffic to processes. This should always be enabled, and effectively disables app settings if disabled.",
```
ppacher, line 20: IMHO the description stutters. What about: `Application specific settings require Process Detection to be enabled` or smth like this?

profile/profile.go
```diff
@@ -166,6 +168,7 @@ func (profile *Profile) parseConfig() error {
 	}
 
 	list, ok := profile.configPerspective.GetAsStringArray(CfgOptionEndpointsKey)
+	profile.endpoints = nil
 	if ok {
```
ppacher, line 172: `nil` is the default value. Is there a reason to set it explicitly?

ppacher (approved):

profile/profile.go
dhaavi: Yes, in case the profile config is changed and reloaded internally. There might be old data assigned.

process/config.go
dhaavi: Thanks, I noted it for later.

PR #277, opened 2021-03-20
Add system resolver compatibility workaround

dhaavi:
I recommend reviewing the commits individually in order to follow the development process.

### DIFF
--- a/network/api.go
+++ b/network/api.go
@@ -11,6 +11,8 @@ import (
 	"github.com/safing/portbase/api"
 	"github.com/safing/portbase/database/query"
 	"github.com/safing/portbase/utils/debug"
+	"github.com/safing/portmaster/network/state"
+	"github.com/safing/portmaster/process"
 	"github.com/safing/portmaster/status"
 )
 
@@ -41,6 +43,18 @@ func registerAPIEndpoints() error {
 				Description: "Specify a query to limit the connections included in the report. The default is to include all connections.",
 			},
 		},
+	}); err != nil {
+		return err
+	}
+
+	if err := api.RegisterEndpoint(api.Endpoint{
+		Path: "debug/network/state",
+		Read: api.PermitUser,
+		StructFunc: func(ar *api.Request) (i interface{}, err error) {
+			return state.GetInfo(), nil
+		},
+		Name:        "Get Network State Table Data",
+		Description: "Returns the current network state tables from the OS.",
 	}); err != nil {
 		return err
 	}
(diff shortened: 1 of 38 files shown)


network/api.go
```diff
@@ -192,7 +208,7 @@ func buildNetworkDebugInfoData(debugConns []*Connection) string {
 
 func (conn *Connection) debugInfoLine() string {
 	var connectionData string
-	if conn.ID != "" {
+	if conn.Type == DNSRequest { // conn.ID !=
```
ppacher: I guess the comment at the end of the line can be removed

network/database.go
```diff
@@ -27,37 +26,86 @@ type StorageInterface struct {
 	storage.InjectBase
 }
 
+// Database prefixes:
+// Processes:       network:tree/<PID>
+// DNS Requests:    network:tree/<PID>/dns/<ID>
+// IP Connections:  network:tree/<PID>/ip/<ID>
```
ppacher, line 32: Before releasing we need to make sure to NOT BREAK the UI. AFAIK this should not be a problem but we must try that first!

network/database.go
```diff
@@ -27,37 +26,86 @@ type StorageInterface struct {
...
+		fallthrough
+	case 2:
+		var err error
+		if segments[1] == "" {
+			pid = process.UndefinedProcessID
+		} else {
+			pid, err = strconv.Atoi(segments[1])
+			if err != nil {
+				return 0, "", "", false
+			}
```
ppacher, line 77: shouldn't we log that error? We expect to be able to parse it as a number so this is likely something that points to a bigger issue

process/profile.go
```diff
@@ -54,6 +55,20 @@ func (p *Process) GetProfile(ctx context.Context) (changed bool, err error) {
...
+		case "windows":
+			if (p.Path == `C:\Windows\System32\svchost.exe` || p.Path == `C:\Windows\system32\svchost.exe`) &&
+				(strings.Contains(p.SpecialDetail, "Dnscache") || strings.Contains(p.CmdLine, "-k NetworkService")) {
+				profileID = profile.SystemResolverProfileID
+			}
+		case "linux":
+			switch p.Path {
+			case "/lib/systemd/systemd-resolved",
+				"/usr/lib/systemd/systemd-resolved":
+				profileID = profile.SystemResolverProfileID
```
ppacher, line 69: For multi-arch systems we should also allow systemd-resolved to be below `/lib64` and `/usr/lib64`.

ppacher (approved):
LGTM. Just some minor changes requested. Did you notice the CI warnings?

network/api.go
dhaavi: Already fixed in previous commit, you seem to have review an old state, before some small fixes.

network/database.go
dhaavi: Yup. Checked.

dhaavi: Ah, wrong comment. This will need some reworking anyway, we'll should have a chat about that.

PR #278, opened 2021-03-29
Fix and improve netenv for Windows and Linux

dhaavi:
Requires https://github.com/safing/portbase/pull/122

### DIFF
--- a/firewall/interception.go
+++ b/firewall/interception.go
@@ -144,6 +144,10 @@ func fastTrackedPermit(pkt packet.Packet) (handled bool) {
 
 		// Always permit ICMP.
 		log.Debugf("filter: fast-track accepting ICMP: %s", pkt)
+
+		// If the packet was submitted to the listener, we must not do a
+		// permanent accept, because then we won't see any future packets of that
+		// connection and thus cannot continue to submit them.
 		if submitted {
 			_ = pkt.Accept()
 		} else {
@@ -157,6 +161,10 @@ func fastTrackedPermit(pkt packet.Packet) (handled bool) {
 
 		// Always permit ICMPv6.
 		log.Debugf("filter: fast-track accepting ICMPv6: %s", pkt)
+
+		// If the packet was submitted to the listener, we must not do a
+		// permanent accept, because then we won't see any future packets of that
+		// connection and thus cannot continue to submit them.
 		if submitted {
 			_ = pkt.Accept()
 		} else {
--- a/firewall/interception.go
+++ b/firewall/interception.go
@@ -139,15 +139,29 @@ func fastTrackedPermit(pkt packet.Packet) (handled bool) {
 
 	switch meta.Protocol {
 	case packet.ICMP:
+		// Submit to ICMP listener.
+		submitted := netenv.SubmitPacketToICMPListener(pkt)
+
 		// Always permit ICMP.
 		log.Debugf("filter: fast-track accepting ICMP: %s", pkt)
-		_ = pkt.PermanentAccept()
+		if submitted {
+			_ = pkt.Accept()
+		} else {
+			_ = pkt.PermanentAccept()
+		}
 		return true
 
 	case packet.ICMPv6:
+		// Submit to ICMP listener.
+		submitted := netenv.SubmitPacketToICMPListener(pkt)
+
 		// Always permit ICMPv6.
 		log.Debugf("filter: fast-track accepting ICMPv6: %s", pkt)
-		_ = pkt.PermanentAccept()
+		if submitted {
+			_ = pkt.Accept()
+		} else {
+			_ = pkt.PermanentAccept()
+		}
 		return true
 
 	case packet.UDP, packet.TCP:
--- a/netenv/icmp_listener.go
+++ b/netenv/icmp_listener.go
@@ -9,12 +9,25 @@ import (
 	"github.com/safing/portmaster/network/packet"
 )
 
+/*
+This ICMP listening system is a simple system for components to listen to ICMP
+packets via the firewall.
+
+The main use case for this is to receive ICMP packets that are not always
+delivered correctly, or need special permissions and or sockets to receive
+them. This is the case when doing a traceroute.
+
+In order to keep it simple, the system is only designed to be used by one
+"user" at at time. Further calls to ListenToICMP will wait for the previous
+operation to complete.
+*/
+
 var (
 	// listenICMPLock locks the ICMP listening system for one user at a time.
 	listenICMPLock sync.Mutex
 
-	// listenICMPEnabled defines whether or not the firewall should send ICMP
-	// packets through this interface.
+	// listenICMPEnabled defines whether or not the firewall should submit ICMP
+	// packets to this interface.
 	listenICMPEnabled = abool.New()
 
 	// listenICMPInput is created for every use of the ICMP listenting system.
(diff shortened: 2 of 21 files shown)


process/profile.go
```diff
@@ -59,7 +59,7 @@ func (p *Process) GetProfile(ctx context.Context) (changed bool, err error) {
 		switch runtime.GOOS {
 		case "windows":
 			if (p.Path == `C:\Windows\System32\svchost.exe` || p.Path == `C:\Windows\system32\svchost.exe`) &&
-				(strings.Contains(p.SpecialDetail, "Dnscache") || strings.Contains(p.CmdLine, "-k NetworkService")) {
+				(strings.Contains(p.SpecialDetail, "Dnscache") || strings.Contains(p.CmdLine, "-s Dnscache")) {
 				profileID = profile.SystemResolverProfileID
```
ppacher, line 63: Is this always the case on all supported windows versions?

network/packet/parse.go
```diff
@@ -137,11 +138,15 @@ func Parse(packetData []byte, pktInfo *Info) error {
...
-		if err := dec(packet, pktInfo); err != nil {
+		if err := dec(packet, pktBase.Info()); err != nil {
 			return err
 		}
 	}
 
+	pktBase.layers = packet
+	if packet.TransportLayer() != nil {
+		pktBase.layer5Data = packet.TransportLayer().LayerPayload()
+	}
```
ppacher, line 149:
```suggestion
	pktBase.layers = packet
	if transport := packet.TransportLayer(); transport != nil {
		pktBase.layer5Data = transport.LayerPayload()
	}
```

netenv/icmp_listener.go
```diff
@@ -0,0 +1,82 @@
...
+	listenICMPLock sync.Mutex
+
+	// listenICMPEnabled defines whether or not the firewall should send ICMP
+	// packets through this interface.
+	listenICMPEnabled = abool.New()
+
+	// listenICMPInput is created for every use of the ICMP listenting system.
+	listenICMPInput     chan packet.Packet
+	listenICMPInputLock sync.Mutex
+)
```
ppacher, line 23: Hm I don't understand that comments. maybe create a file-level comment describing what a "ICMP listening system" is and why it is "per user".
Also, the comment in `listenICMPEnabled` talks abound sending ICMP packets. This seems wrong.

netenv/icmp_listener.go
```diff
@@ -0,0 +1,82 @@
...
+
+func ListenToICMP() (packets chan packet.Packet, done func()) {
+	// Lock for single use.
+	listenICMPLock.Lock()
+
```
ppacher, line 28: A comment on what this method does or how it is meant to be used would be great.
Maybe consider using a `context.Context` for cancellation instead of a `done` function.

netenv/icmp_listener.go
```diff
@@ -0,0 +1,82 @@
...
+	listenICMPInputLock.Lock()
+	defer listenICMPInputLock.Unlock()
+
+	// Check if still enabled.
+	if !listenICMPEnabled.IsSet() {
+		return
+	}
+
+	log.Criticalf("netenv: recvd ICMP packet: %s", pkt)
+
```
ppacher, line 75: I guess this should not be critical

netenv/icmp_listener.go
```diff
@@ -0,0 +1,82 @@
...
+func SubmitPacketToICMPListener(pkt packet.Packet) (submitted bool) {
+	// Hot path.
+	if !listenICMPEnabled.IsSet() {
+		return false
+	}
+
+	// Slow path.
+	submitPacketToICMPListenerSlow(pkt)
+	return true
+}
```
ppacher, line 56: It might not get submitted if the packet channel is full. I guess we should return that here?

firewall/interception.go
```diff
@@ -139,15 +139,29 @@ func fastTrackedPermit(pkt packet.Packet) (handled bool) {
...
+
 		// Always permit ICMP.
 		log.Debugf("filter: fast-track accepting ICMP: %s", pkt)
-		_ = pkt.PermanentAccept()
+		if submitted {
+			_ = pkt.Accept()
+		} else {
+			_ = pkt.PermanentAccept()
+		}
 		return true
```
ppacher, line 152: Can we explain why we only `Accept` the ICMP packet if we were able to submit it to someone listening? I would expect it to be the other way around (but I don't yet know what's that for ...)

netenv/icmp_listener.go
```diff
@@ -0,0 +1,82 @@
...
+	listenICMPEnabled.Set()
+	listenICMPInputLock.Unlock()
+
+	return listenICMPInput, func() {
+		// Close input channel.
+		listenICMPInputLock.Lock()
+		listenICMPEnabled.UnSet()
+		close(listenICMPInput)
+		listenICMPInputLock.Unlock()
+
```
ppacher, line 41: defer the unlock because if `close(listenICMPInput)` panics (because someone already called close on it) we wouldn't be left over with a stale and locked mutex. In case the panic gets recovered (which it should) the ICMP listener can still be reused.

netenv/location.go
```diff
@@ -2,86 +2,265 @@ package netenv
...
 
-	// First check if we have an assigned IPv6 address. Return that if available.
-	globalIPv4, _, err := GetAssignedGlobalAddresses()
+	// Find best location.
+	var best *DeviceLocation
+	for _, dl := range locations.All {
+		if best == nil || dl.SourceAccuracy > best.SourceAccuracy {
+			best = dl
+		}
+	}
```
ppacher, line 128: Instead of checking `best == nil` you can just start with `best := loc`.

netenv/location.go
```diff
@@ -2,86 +2,265 @@ package netenv
...
-		log.Warningf("netenv: location approximation: failed to get assigned global addresses: %s", err)
-	} else if len(globalIPv4) > 0 {
-		return globalIPv4[0], nil
+		log.Warningf("netenv: failed to get geolocation data of %s (from %s): %s", ip, source, err)
+	} else {
+		loc.Continent = geoLoc.Continent.Code
+		loc.Country = geoLoc.Country.ISOCode
+		loc.ASN = geoLoc.AutonomousSystemNumber
+		loc.ASOrg = geoLoc.AutonomousSystemOrganization
+	}
```
ppacher, line 140: This is not related to the `locations` variable and can be fetched outside the `locationsLock`.

netenv/location.go
```diff
@@ -91,72 +270,106 @@ next:
...
+				case layers.ICMPv4TypeDestinationUnreachable:
+					// We have received a valid destination unreachable response, abort.
+					return false
+
+				case layers.ICMPv4TypeTimeExceeded:
+					// We have received a valid time exceeded error.
+					// If message came from a global unicast, us it!
+					if netutils.GetIPScope(remoteIP) == netutils.Global {
+						return SetInternetLocation(remoteIP, SourceTraceroute)
+					}
```
ppacher, line 336: That check is done in SetInternetLocation as well so we can just call it and abort if it returns true.

ppacher (approved):
Great work! Looking forward to use that! Just a few minor things noted.

netenv/icmp_listener.go
dhaavi: After thinking about it, I'd like to talk about that again, leaving at it is for now.

PR #285, opened 2021-04-07
Fix early connection handling and mimetypes

dhaavi:


firewall/interception.go
```diff
@@ -103,20 +104,60 @@ func handlePacket(ctx context.Context, pkt packet.Packet) {
...
+	switch {
+	case created && shared:
+		log.Tracer(pkt.Ctx()).Tracef("filter: created new connection %s (shared)", conn.ID)
+	case created:
+		log.Tracer(pkt.Ctx()).Tracef("filter: created new connection %s", conn.ID)
+	case shared:
+		log.Tracer(pkt.Ctx()).Tracef("filter: assigned connection %s (shared)", conn.ID)
+	default:
+		log.Tracer(pkt.Ctx()).Tracef("filter: assigned connection %s", conn.ID)
+	}
```
ppacher, line 156:
```suggestion
	sharedStr := ""
	if shared {
		sharedStr = " (shared)"
	}
	if created {
		log.Tracer(pkt.Ctx()).Tracef("filter: created new connection %s%s", conn.ID, shared)
	} else {
		log.Tracer(pkt.Ctx()).Tracef("filter: assigned connection %s%s", conn.ID, shared)
	}
```

ppacher (commented):
lgtm

PR #296, opened 2021-04-29
Add OpenBSD osIdentifier

qbit:
This lets portmaster build on OpenBSD.

### DIFF
--- a/profile/fingerprint/const.go
+++ b/profile/fingerprint/const.go
@@ -5,4 +5,5 @@ const (
 	PlatformLinux   = "linux"
 	PlatformWindows = "windows"
 	PlatformMac     = "macos"
+	PlatformOpenBSD = "openbsd"
 )
--- a/profile/fingerprint/const_openbsd.go
+++ b/profile/fingerprint/const_openbsd.go
@@ -1,0 +1,6 @@
+package profile
+
+// OS Identifier
+const (
+	osIdentifier = PlatformOpenBSD
+)


qbit:
I forgot to mention that this also needs: https://github.com/safing/portbase/pull/125

I used `replace github.com/safing/portbase => ../portbase` in `go.mod` to test this MR.

dhaavi:
Hey @qbit, thanks for your contributions.

I will take a closer look at this next week. In the mean time, can you shed more light on what you are trying to achieve?

Making the Portmaster able to compile on OpenBSD is a good step forward to support OpenBSD, but as far as I understand, there will be a lot more work necessary in order to make the Portmaster compatible with OpenBSD.

When I have it right in my head, the only thing that'll work right now is the internal DNS server. Both the firewall and the process detection won't work.

qbit:
> In the mean time, can you shed more light on what you are trying to achieve?

Sure, end goal is to make everything functional. gopsutil recently gained support for OpenBSD/arm64 and already has support for OpenBSD/amd64. I haven't looked into the firewall bits yet, but I plan on doing so.

dhaavi:
> Sure, end goal is to make everything functional. gopsutil [recently](https://github.com/shirou/gopsutil/commit/992d17af9403179c3a9bc9b1f45af98e9be6e4a1) gained support for OpenBSD/arm64 and already has support for OpenBSD/amd64. I haven't looked into the firewall bits yet, but I plan on doing so.

That is great!
You can find the two primary OS integration points here:

1. [Packet Interception](https://github.com/safing/portmaster/blob/develop/firewall/interception/interception_default.go)
  - The packet input channel should ingest all packets.
  - It's not mandatory, but I'd highly advise to have persistence (see [Packet interface](https://github.com/safing/portmaster/blob/a3b495f6c921d4189a57edda10c55cb71ef37f86/network/packet/packet.go#L221)) on the OS side of the integration. You can look at the [iptables rules](https://github.com/safing/portmaster/blob/develop/firewall/interception/nfqueue_linux.go) for an example.
2. [Packet to Process Attribution](https://github.com/safing/portmaster/blob/a3b495f6c921d4189a57edda10c55cb71ef37f86/network/state/system_default.go)
  - This uses `/proc` on Linux and the `iphlpapi.dll` on Windows. I guess OpenBSD should be able to fit this interface.

dhaavi:
Oh and, before I forget:

Please read our [contribution guide](https://docs.safing.io/portmaster/guides/contribute) - with such a big feature please discuss the concept with us in an issue before starting implementation. We don't want your precious time to go to waste!

dhaavi (approved):

PR #299, opened 2021-05-04
Improvements for notifications and events

dhaavi:
and other things

Requires https://github.com/safing/portbase/pull/126.


ppacher (dismissed):

intel/filterlists/module.go
```diff
@@ -104,3 +102,11 @@ func stop() error {
 	filterListsLoaded = make(chan struct{})
 	return nil
 }
+
+func warnAboutDisabledFilterLists() {
+	module.Warning(
+		filterlistsDisabled,
+		"Filter Lists Are Initializing",
+		"The Filter Lists are waiting for update system to check for updates in order to retrieve the filter list data. Until the filter lists are fully initialized, the filter lists feature is disabled.",
+	)
```
ppacher, line 111: "Filter Lists" is used quite often here. Maybe change that to something like: "Filter lists are being downloaded and processed in the background. Until then the filter lists feature is disabled".

intel/filterlists/updater.go
```diff
@@ -23,17 +23,19 @@ func tryListUpdate(ctx context.Context) error {
...
 		if !isLoaded() {
-			module.Error(filterlistsDisabled, err.Error())
+			warnAboutDisabledFilterLists()
 		} else {
-			module.Warning(filterlistsUpdateFailed, err.Error())
+			module.Warning(
+				filterlistsUpdateFailed,
+				"Filter Lists Update Failed",
+				fmt.Sprintf("The filter lists system failed to process an update. Depending on the previous state, the filtering capabilities are now either impaired or not given. Refer to the error for more details: %s", err.Error()),
+			)
```
ppacher, line 34: I'm not sure if talking about the "previous state" this is too technical. Also, I would remove the "Refer to the error" part as that makes only sense when reading the source code and would just do "Error: "

intel/filterlists/updater.go
```diff
@@ -44,7 +46,11 @@ func performUpdate(ctx context.Context) error {
 	}
 	defer updateInProgress.UnSet()
 
-	module.Hint(filterlistsUpdateInProgress, filterlistsUpdateInProgressDescr)
+	module.Hint(
+		filterlistsUpdateInProgress,
+		"Filter Lists Update In Progress",
+		"The filter list system is processing updates. While this might slightly degrade performance, the filter list system stays functional during this process.",
+	)
```
ppacher: I would not talk about the "filter list system" as users don't know about internal systems/boundaries/modules. I'd just say: "The Portmaster is processing a filterlist update. While this might slightly degrade performance, the Portmaster is still protecting your privacy "

intel/filterlists/updater.go
```diff
@@ -119,7 +125,11 @@ func performUpdate(ctx context.Context) error {
 			// if we failed to remove all stale cache entries
 			// we abort now WITHOUT updating the database version. This means
 			// we'll try again during the next update.
-			module.Warning(filterlistsStaleDataSurvived, filterlistsStaleDataDescr)
+			module.Warning(
+				filterlistsStaleDataSurvived,
+				"Filter Lists May Overblock",
+				fmt.Sprintf("The filter lists system successfully applied an update, but failed to delete old data. This means that the filtering system performs as usual, but may block more entries than it should. Refer to the error for more details: %s", err.Error()),
+			)
```
ppacher, line 127: Same as above

profile/config-update.go
```diff
@@ -123,7 +123,8 @@ func updateGlobalConfigProfile(ctx context.Context, task *modules.Task) error {
 		// Add module warning to inform user.
 		module.Warning(
 			globalConfigProfileErrorID,
-			fmt.Sprintf("Failed to process global settings: %s", err),
+			"Internal Settings Failure",
+			fmt.Sprintf("The app settings layering system failed to process the global settings. This means that some global settings might not be applied correctly. You can try restarting the Portmaster to resolve this problem. Refer to the error for more details: %s", err),
 		)
```
ppacher, line 128: As with the "filter list system", I would not talk about "app settings layering system". That's way to complex and our users will feel completely overwhelmed by this message. I'd remove the first sentence completely and just say: "Some global settings might not be applied correctly. You can try restarting the Portmaster to resolve this problem. Error: %s"

updates/config.go
```diff
@@ -102,20 +104,33 @@ func updateRegistryConfig(_ context.Context, _ interface{}) error {
...
+			if err := TriggerUpdate(); err != nil {
+				log.Warningf("updates: failed to trigger update: %s", err)
+			}
+			log.Infof("updates: automatic updates are now enabled")
+		} else {
+			notifications.NotifyWarn(
+				updatesDisabledNotificationID,
+				"Automatic Updates Disabled",
+				"The automatic update system is disabled through configuration. Please note that this is potentially dangerous, as this also affects security updates as well as the filter lists and threat intelligence feeds.",
+				notifications.Action{
```
ppacher, line 124: I'd say: "Automatic updates are disabled through configuration"

profile/special.go
```diff
@@ -117,3 +138,63 @@ func getSpecialProfile(profileID, linkedPath string) *Profile {
...
+// check if the profile is outdated and can be upgraded.
+func specialProfileNeedsReset(profile *Profile) bool {
+	switch {
+	case profile.Source != SourceLocal:
+		// Special profiles live in the local scope only.
+		return false
+	case profile.LastEdited > 0:
+		// Profile was edited - don't override user settings.
+		return false
+	}
```
ppacher, line 154: Should we create a notification for that case so the user knows that we updated the defaults?

ppacher (approved):

ppacher (approved):

PR #327, opened 2021-06-01
Improve system resolver and ICMP handling

dhaavi:


firewall/interception.go
```diff
@@ -184,30 +185,39 @@ func fastTrackedPermit(pkt packet.Packet) (handled bool) {
...
-			_ = pkt.Accept()
-		} else {
+	case packet.ICMP, packet.ICMPv6:
+		// Load packet data.
+		err := pkt.LoadPacketData()
+		if err != nil {
+			log.Debugf("filter: failed to load ICMP packet data: %s", err)
 			_ = pkt.PermanentAccept()
+			return true
+		}
```
ppacher, line 195: should we really do a permanent accept here? and only log on warning level?

ppacher (approved):

PR #331, opened 2021-06-03
Improve release channels

dhaavi:


cmds/portmaster-start/lock.go
```diff
@@ -86,7 +85,7 @@ func createInstanceLock(lockFilePath string) error {
 	}
 
 	// create lock file
-	err = ioutil.WriteFile(lockFilePath, []byte(fmt.Sprintf("%d", os.Getpid())), 0666)
+	err = ioutil.WriteFile(lockFilePath, []byte(fmt.Sprintf("%d", os.Getpid())), 0666) //nolint:gosec
 	if err != nil {
```
ppacher, line 89: That permission is actually insecure. why not using something like 0600 or 0655?

cmds/portmaster-start/logs.go
```diff
@@ -96,7 +96,7 @@ func getPmStartLogFile(ext string) *os.File {
 	}, info.Version(), ext)
 }
 
-//nolint:deadcode,unused // false positive on linux, currently used by windows only
+//nolint:unused // false positive on linux, currently used by windows only
 func logControlError(cErr error) {
 	// check if error present
 	if cErr == nil {
```
ppacher, line 102: we cloud move it to a `_windows.go` file then

updates/helper/updates.go
```diff
@@ -0,0 +1,68 @@
...
+		// User interface components
+		PlatformIdentifier("app/portmaster-app.zip"),
+		"all/ui/modules/portmaster.zip",
+		"all/ui/modules/assets.zip",
+
+		// Filter lists data
+		"all/intel/lists/base.dsdl",
+		"all/intel/lists/intermediate.dsdl",
+		"all/intel/lists/urgent.dsdl",
+
```
ppacher, line 54: what if one of those resources would not exist because it's not released yet? (that would only happen with a self-hosted update-server though ...)

ppacher (approved):

cmds/portmaster-start/lock.go
dhaavi: I'll add a TODO. Don't wan't add more changes to this right now.

updates/helper/updates.go
dhaavi: Good point. I think the update process would just fail. I'll add a TODO to investigate in the future.

cmds/portmaster-start/logs.go
dhaavi: Good point. I'll need to have to look into that more and understand it again. ;)
Will add a TODO tough.

ppacher (approved):

PR #347, opened 2021-06-24
Rename special release channel to support

dhaavi:


updates/config.go
```diff
@@ -50,15 +50,21 @@ func registerConfig() error {
...
 			},
 			{
 				Name:        "Staging",
 				Description: "Dangerous development releases for testing random things and experimenting. Only use temporarily and when instructed.",
 				Value:       helper.ReleaseChannelStaging,
 			},
+			// TODO: Remove as soon as everyone has switched away.
+			{
+				Name:        "Special (Deprecated!)",
+				Description: "This channel has been deprecated. If selected, the Stable channel will be used instead.",
```
ppacher, line 65: Who does guarantee that? Or ist stable Just a fallback for invalid values

ppacher (approved):

updates/config.go
dhaavi: Yes, stable is fallback.

PR #366, opened 2021-08-07
Add note about disconnects

boehs:
Happened to me /shrug

### DIFF
--- a/intel/filterlists/module.go
+++ b/intel/filterlists/module.go
@@ -107,6 +107,6 @@ func warnAboutDisabledFilterLists() {
 	module.Warning(
 		filterlistsDisabled,
 		"Filter Lists Are Initializing",
-		"Filter lists are being downloaded and set up in the background. They will be activated as configured when finished.",
+		"Filter lists are being downloaded and set up in the background. They will be activated as configured when finished. During this process you may lose access to the internet.",
 	)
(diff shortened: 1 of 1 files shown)


davegson:
Hi @boehs, thanks for the pull request. Having the Internet cut off while setting up the filter lists is definitely unintended behavior, sorry this happened!

However, I am unhappy solving this with an extensive warning message. I feel our strong branding of Portmaster being in "alpha" already communicates these kind of hick-ups can occur.

So instead of adding a warning, I would much rather get to the root of this issue and fix it, rendering such a message superfluous. Would you agree with this take? Would it be possible to [open up a bug report](https://github.com/safing/portmaster/issues/new?assignees=&labels=bug&template=report-bug.md) when this happens again?

davegson:
closing for now, however you are welcome to comment and advocate for a different strategy

davegson:
as an update, this seems to happen more often than I assumed:
https://old.reddit.com/r/safing/comments/p31gr3/portmaster_doesnt_load_and_cuts_off_connection/

did it also completely shut off the Internet for you until you uninstalled/stopped the Portmaster?

boehs:
Sorry, was on a long trip. Do you need anything anymore? It seems the reddit thread was useful

PR #375, opened 2021-08-18
Improve geoip database downloading and do not block connections

ppacher:


intel/geoip/database.go
```diff
@@ -13,106 +13,205 @@ import (
...
+// updateBroadcaster stores a geoIPDB and provides synchronized
+// access to the MMDB reader. It also supports broadcasting to
+// multiple waiters when a new database becomes available.
+type updateBroadcaster struct {
+	rw sync.RWMutex
+	db *geoIPDB
 
-	dbLock.Lock()
-	defer dbLock.Unlock()
+	waiters []chan struct{}
```
dhaavi, line 41: Wouldn't a single wait channel do it too? Access to it is locked anyway, so you can just close it and replace it with a new one afterwards.

intel/geoip/database.go
```diff
@@ -13,106 +13,205 @@ import (
...
-	if err != nil {
-		return fmt.Errorf("could not get GeoIP v6 database file: %s", err)
+func (upd *updateWorker) start() {
+	upd.once.Do(func() {
+		module.StartServiceWorker("geoip-updater", time.Second*10, upd.run)
+	})
+}
+
+func (upd *updateWorker) run(ctx context.Context) error {
+	firstTime := true
```
dhaavi, line 152: Instead of using a "first" flag, you can just move to select block to the end of the for loop.

intel/geoip/database.go
```diff
@@ -13,106 +13,205 @@ import (
...
+			}
+		}
+		ub.rw.Unlock()
+		return nil
 	}
-	geoDBv4Reader, err = maxminddb.Open(unpackedV4)
-	if err != nil {
-		return err
+	rd := ub.db.Reader
+	ub.rw.Unlock()
```
dhaavi, line 130: I'd really like to have the unlocks in a defer, so we don't deadlock ourselves.

Is a getter method for `ub.db` and locking for `getWaiter` a possibility?

dhaavi (changes requested):
Much cleaner and stable now! Great!

Some small improvement suggestions.

intel/geoip/database.go
ppacher: Not here because if I unlock a pending call to `ub.ReplaceDatabase()` might occur that we would miss then. I also like the defered ones more but I need to keep the lock locked :)

intel/geoip/database.go
ppacher: yeah, I moved it to the top because I had two `continue` statements in early versions but they are not there anymore so I can move it to the bottom again. Thanks for pointing out

intel/geoip/database.go
ppacher: good point :)

intel/geoip/database.go
dhaavi: Not exactly beautiful, but you could do the waiting itself in a defer that you define before the unlock defer:

Use a named return value of `r` and just return nil with `wait = true`:
```go
    defer func() {
        if wait {
	    select {
	    case <-waiter:
		    // call this method again but this time we don't allow
		    // it to wait since there must be a open database anyway ...
		    r = upd.GetReader(v6, false)
	    case <-time.After(time.Second):
		    // we tried hard but failed so give up here
	    }
	}
}()
```

dhaavi: As the code in question currently cannot even panic, let's leave it as it is.

PR #379, opened 2021-08-19
Improve endpoint/rule lists and filtering of DNS requests

dhaavi:


ppacher (dismissed):

firewall/master.go
```diff
@@ -99,25 +92,18 @@ func DecideOnConnection(ctx context.Context, conn *network.Connection, pkt packe
...
 	done, defaultAction := runDeciders(ctx, defaultDeciders, conn, layeredProfile, pkt)
 	if done {
 		return
 	}
 
+	// DNS Request are always default allowed, as the endpoint lists could not
+	// be checked fully.
+	if conn.Type == network.DNSRequest {
+		conn.Accept("allowing dns request", noReasonOptionKey)
+	}
```
ppacher, line 105: aren't we missing a `return` here?

profile/config.go
```diff
@@ -182,6 +182,8 @@ func registerConfiguration() error {
 Additionally, you may supply a protocol and port just behind that using numbers ("6/80") or names ("TCP/HTTP").  
 In this case the rule is only matched if the protocol and port also match.  
 Example: "192.168.0.1 TCP/HTTP"
+
+Important: DNS Requests are only matched against domain and filter list rules, all others require an IP address and are checked only with the ensuing IP connection.
 `, `"`, "`")
```
ppacher, line 187: IMHO `ensuing` is not a commonly used word. Maybe we should rephrase that for non-nativ-english speakers.

ppacher (changes requested):

firewall/master.go
dhaavi: Very good catch!

PR #399, opened 2021-09-26
Updates and Improvements (related to SPN v0.3)

dhaavi:
And some other small things.
Would recommend to review commits individually.

Internal dependencies have not yet been updated. Build in local dev environment with updated repos.

### DIFF
--- a/network/connection.go
+++ b/network/connection.go
@@ -502,13 +502,14 @@ func (conn *Connection) SetVerdict(newVerdict Verdict, reason, reasonOptionKey s
 		conn.Verdict = newVerdict
 		conn.Reason.Msg = reason
 		conn.Reason.Context = reasonCtx
+
+		conn.Reason.Profile = ""
+		conn.Reason.OptionKey = ""
 		if reasonOptionKey != "" && conn.Process() != nil {
-			conn.Reason.OptionKey = reasonOptionKey
 			conn.Reason.Profile = conn.Process().Profile().GetProfileSource(conn.Reason.OptionKey)
-		} else {
-			conn.Reason.OptionKey = ""
-			conn.Reason.Profile = ""
+			conn.Reason.OptionKey = reasonOptionKey
 		}
+
 		return true
 	}
 	return false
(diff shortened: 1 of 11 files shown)


network/connection.go
```diff
@@ -505,6 +505,9 @@ func (conn *Connection) SetVerdict(newVerdict Verdict, reason, reasonOptionKey s
 		if reasonOptionKey != "" && conn.Process() != nil {
 			conn.Reason.OptionKey = reasonOptionKey
 			conn.Reason.Profile = conn.Process().Profile().GetProfileSource(conn.Reason.OptionKey)
+		} else {
+			conn.Reason.OptionKey = ""
+			conn.Reason.Profile = ""
 		}
```
ppacher, line 511:
```suggestion
              conn.Reason.OptionKey = reasonOptionKey
              conn.Reason.Profile = ""
		if process := conn.Process(); process != nil {
			conn.Reason.Profile = process.Profile().GetProfileSource(conn.Reason.OptionKey)
		}
```

ppacher: Though, I'm not sure `conn.Process()` can even return `nil`. Did you meant to check `conn.Process().Profile()` for `nil`?

updates/helper/updates.go
```diff
@@ -20,6 +27,26 @@ func PlatformIdentifier(identifier string) string {
...
+	)
+
+	// Stop here if we only want intel data.
+	if intelOnly.IsSet() {
+		return
+	}
+
 	// Binaries
 	if onWindows {
 		identifiers = []string{
```
ppacher, line 52: That must be an `append()` now! Otherwise if `!intelOnliy.IsSet()` than intel data is not part of the identifiers anymore!

ppacher (changes requested):

resolver/resolver-tcp.go
```diff
@@ -189,7 +189,7 @@ func (tr *TCPResolver) Query(ctx context.Context, q *Query) (*RRCache, error) {
 	select {
 	case resolverConn.queries <- tq:
 	case <-ctx.Done():
-		return nil, ErrCancelled
+		return nil, context.Canceled
```
ppacher:
```suggestion
case <-ctx.Done():
    return nil, ctx.Err() // ctx.Err()  must either be context.Canceled or context.DeadlineExceeded
```

ppacher (dismissed):

PR #401, opened 2021-09-29
Fill missing location info based on existing info

dhaavi:


intel/geoip/fill_missing.go
```diff
@@ -0,0 +1,261 @@
...
+		l.Country.ISOCode != "" {
+		if c, ok := countryCoordinates[l.Country.ISOCode]; ok {
+			l.Coordinates = c
+		}
+	}
+}
+
+var countryCoordinates = map[string]Coordinates{
+	"AD": {500, 42, 1},
+	"AE": {500, 23, 53},
```
ppacher, line 18: Is that something we should distribute as "intel" data?

ppacher (approved):

PR #404, opened 2021-10-04
Improve updates, netenv location and connection logging

dhaavi:


netenv/location.go
```diff
@@ -100,15 +100,26 @@ func (dl *DeviceLocation) IsMoreAccurateThan(other *DeviceLocation) bool {
...
-	case dl.Location.Country.ISOCode != "" && other.Location.Country.ISOCode == "":
+	case dl.Location.Country.ISOCode != "" &&
+		other.Location.Country.ISOCode == "":
 		// Having a Country is better than having none.
 		return true
+	case (dl.Location.Coordinates.Latitude != 0 ||
+		dl.Location.Coordinates.Longitude != 0) &&
+		other.Location.Coordinates.Latitude == 0 &&
+		other.Location.Coordinates.Longitude == 0:
+		// Having Coordinates is better than having none.
```
ppacher, line 122: I guess you're missing a `return true` here?

netenv/online-status.go
```diff
@@ -200,13 +200,18 @@ func updateOnlineStatus(status OnlineStatus, portalURL *url.URL, comment string)
...
 		} else {
 			log.Infof("netenv: setting online status to %s (%s)", status, comment)
 		}
 		triggerNetworkChangeCheck()
+
+		// Trigger update check when coming (semi) online.
+		if Online() {
+			_ = updates.TriggerUpdate(false)
+		}
 	}
```
ppacher, line 215: Hm, I guess it depends but should'nt the updates module better listen to `OnlineStatusChangedEvent` instead of adding a new dependency on `updates` to `netenv`?

network/connection.go
```diff
@@ -615,8 +615,11 @@ func (conn *Connection) packetHandler() {
 		} else {
 			defaultFirewallHandler(conn, pkt)
 		}
+
 		// log verdict
 		log.Tracer(pkt.Ctx()).Infof("filter: connection %s %s: %s", conn, conn.Verdict.Verb(), conn.Reason.Msg)
+		// submit trace logs
+		log.Tracer(pkt.Ctx()).Submit()
```
ppacher, line 622: I guess this could be `defer`ed

ppacher (changes requested):

netenv/online-status.go
dhaavi: I tried listening to the event, but cannot get the online status, because of a dependency loop. Because the `OnlineStatus` is a separate type, I cannot even manually cast it when it is supplied as event data.

network/connection.go
dhaavi: Submitting itself is not expensive and must happen earlier so that the logs are in the correct order. Before, the SPN integration would log before the connection got accepted.

ppacher (approved):

PR #405, opened 2021-10-07
Reduce open dns requests by better handling dual stack requests

dhaavi:


network/dns.go
```diff
@@ -29,33 +29,38 @@ const (
...
 
-	key := getDNSRequestCacheKey(pid, fqdn)
-	_, ok := openDNSRequests[key]
-	if ok {
-		delete(openDNSRequests, key)
-		return
-	}
+	// Delete PID-specific requests.
+	delete(openDNSRequests, getDNSRequestCacheKey(pid, fqdn, dns.TypeA))
+	delete(openDNSRequests, getDNSRequestCacheKey(pid, fqdn, dns.TypeAAAA))
```
ppacher, line 42: Shouldn't we only remove the DNS request for which we recieved the response? We now delete all `A` and `AAAA` requests for the domain which might cause the next answer  received to not have a open-DNS request available

network/dns.go
```diff
@@ -29,33 +29,38 @@ const (
...
 
-	key := getDNSRequestCacheKey(conn.process.Pid, conn.Entity.Domain)
+	// Only save open A and AAAA requests.
+	switch qType {
+	case dns.TypeA, dns.TypeAAAA:
+	default:
+		return
+	}
+
+	key := getDNSRequestCacheKey(conn.process.Pid, conn.Entity.Domain, qType)
```
ppacher, line 63: Why don't we save other request types as well anymore?

network/dns.go
```diff
@@ -29,43 +30,56 @@ const (
...
+
+	// Try to "merge" A and AAAA requests into the resulting connection.
+	// Save others immediately.
+	switch uint16(q.QType) {
+	case dns.TypeA, dns.TypeAAAA:
+	default:
+		conn.Save()
+		return
+	}
+
```
ppacher, line 67: We should save the query type and other query metadata (maybe the whole query) along the connection. Maybe as a nilable `DNSContext` or similar. Otherwise we won't be able to display any meaningful information about DNS requests except the requested domain ...

ppacher (approved):

PR #413, opened 2021-10-13
Add and fix port range rules and update documentation for it

### LINKED ISSUE #412: Firewall port ranges (+general firewall rules guide?)
**What would you like to add or change?**:
I would like to be able to specify an incoming/ougoing port range in the application rules.

**Why do you and others need this?**:
Software like [KDE connect](https://kdeconnect.kde.org/) requires the firewall to allow requests from a specific port range for the syncing to happen. Specifically the ports `1714-1764` for both tcp and udp ([Source](https://www.incredigeek.com/home/allow-kde-connect-through-firewall/)).
I also cannot whitelist a specific local IP address because it might change once the lease runs out.

While researching on writing these rules, I could not find a guide in the docs on even the basic structure of a rule
https://docs.safing.io/portmaster/architecture/core-service/core/#firewall < Not there
https://docs.safing.io/portmaster/architecture/core-service/privacy-filter/#rules < Or here

And I could not find a valid syntax even with trial and error
```
* tcp/1714-1764
* tcp/1714:1764
* tcp/1714 : 1764
and some more...
```
All of them failing with an error message
image

I even went so far as to try and dissect the validation regex with still no luck

So I guess the question is: Is it possible to specify a port range in firewall rules? If so, could it (together with a general syntax guide) be added to the docs?

dhaavi:
Resolves #412.

### DIFF
--- a/profile/config.go
+++ b/profile/config.go
@@ -182,11 +182,23 @@ func registerConfiguration() error {
 - Match anything: "*"
 
 Additionally, you may supply a protocol and port just behind that using numbers ("6/80") or names ("TCP/HTTP").  
-In this case the rule is only matched if the protocol and port also match.  
-Example: "192.168.0.1 TCP/HTTP"
+Port ranges are defined by using a hyphen ("TCP/1-1024"). Omit the port to match any.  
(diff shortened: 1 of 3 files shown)


ppacher (approved):
Lgtm

PR #453, opened 2021-11-18
Improve nameserver

dhaavi:
- Resolve hostname to localhost
- Add OpenSuse connectivity domain
- Ignore un-parse-able RR records when serializing
- Add nscd to system resolver detection

### DIFF
--- a/nameserver/nameserver.go
+++ b/nameserver/nameserver.go
@@ -19,6 +19,8 @@ import (
 	"github.com/miekg/dns"
 )
 
+var hostname string
+
 func handleRequestAsWorker(w dns.ResponseWriter, query *dns.Msg) {
 	err := module.RunWorker("dns request", func(ctx context.Context) error {
 		return handleRequest(ctx, w, query)
@@ -87,8 +89,8 @@ func handleRequest(ctx context.Context, w dns.ResponseWriter, request *dns.Msg) 
 		return reply(nsutil.Refused("unsupported qclass"))
 	}
 
-	// Handle request for localhost.
-	if strings.HasSuffix(q.FQDN, "localhost.") {
+	// Handle request for localhost and the hostname.
+	if strings.HasSuffix(q.FQDN, "localhost.") || q.FQDN == hostname {
 		tracer.Tracef("nameserver: returning localhost records")
 		return reply(nsutil.Localhost())
 	}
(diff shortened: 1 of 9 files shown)


nameserver/nameserver.go
```diff
@@ -87,8 +89,8 @@ func handleRequest(ctx context.Context, w dns.ResponseWriter, request *dns.Msg)
 		return reply(nsutil.Refused("unsupported qclass"))
 	}
 
-	// Handle request for localhost.
-	if strings.HasSuffix(q.FQDN, "localhost.") {
+	// Handle request for localhost and the hostname.
+	if strings.HasSuffix(q.FQDN, "localhost.") || q.FQDN == hostname {
 		tracer.Tracef("nameserver: returning localhost records")
 		return reply(nsutil.Localhost())
 	}
```
ppacher, line 96: The `hostname` is often resolved to a different IP address like `´127.0.1.1` or a static IP assigned to the primary interface. I guess if we want to handle that correctly we should finally start parsing `/etc/hosts` on linux and use that.

ppacher (dismissed):

nameserver/nameserver.go
dhaavi: Afaict, the `127.0.1.1` is used by the OS to do a IP -> hostname mapping, not necessarily for resolving.
I don't think this should be a problem, and if it does, we'll handle it then.
Right now, it's an improvement because some tools (eg. sudo) do want to resolve the hostname via DNS.

process/profile.go
```diff
@@ -72,7 +72,9 @@ func (p *Process) GetProfile(ctx context.Context) (changed bool, err error) {
 			case "/lib/systemd/systemd-resolved",
 				"/usr/lib/systemd/systemd-resolved",
 				"/lib64/systemd/systemd-resolved",
-				"/usr/lib64/systemd/systemd-resolved":
+				"/usr/lib64/systemd/systemd-resolved",
+				"/usr/sbin/nscd",
+				"/usr/local/sbin/nscd":
```
ppacher, line 77: I'm not sure we should consider binaries in `/usr/local` here. No package manager will install something there users that install software by compiling and (likely) `make install`ing them can also create symlinks for that. We might open up some security issues if we consider `/usr/local` as well.

ppacher (approved):

ppacher (approved):

PR #460, opened 2021-12-02
Make own IP check more resistant

dhaavi:


netenv/addresses.go
```diff
@@ -70,18 +73,33 @@ func IsMyIP(ip net.IP) (yes bool, err error) {
...
+
 	// Refresh assigned networks.
 	interfaceNetworks, err := net.InterfaceAddrs()
 	if err != nil {
+		// Save error for one second.
+		// In some cases the system blocks on this call, which piles up to
+		// literally over thousand goroutines wanting to try this again.
+		myNetworksRefreshError = err
+		myNetworksRefreshFailingUntil = time.Now().Add(1 * time.Second)
 		return false, fmt.Errorf("failed to refresh interface addresses: %s", err)
```
ppacher, line 103: We could also use the "singleflight" package here and use the network interface name as the key. Otherwise LGTM

ppacher (approved):

Issue #391, opened 2021-09-10
App Settings / Profiles have interference from Prompts, Metadata Update


dhaavi:
**What happened**:

We've had multiple reports within the last months where settings were not saved, disappeared or were not added at all. These related issues may or may not be related and may or may not be fully resolved: #260, #312, #368, https://github.com/safing/portmaster-ui/issues/118

None of these had a particularly clean solution. I suspect that there is still some interference when saving profiles, or that profiles get "out of sync" so they are not saved and/or applied correctly.

**What did you expect to happen?**:

All settings should be preserved, always.

**How did you reproduce it?**:

We were not yet able to reproduce.

A user reported that this might be able to reproduce it:
1. use main app settings to: *allow domains* / *block domains*
2. use the left panel to: *allow domains* / *block domains*
3. Go back to the main app settings and all outgoing settings are wiped!

northys:
All I can say is that the "didn't save" case never happened to me. It always disappears after a while. I whitelist analytics.google.com (which is blocked by ublock anyway) at least twice a month because I sometimes have to check the stats for client websites. Before I had troubles with accessing AWS console because they load CSS and JS from random cloudflare domain that portmaster do not like so I used whitelist for this domain but it always vanished after a while.

Please try the following - whitelist analytics.google.com (or any other domain) in this cute drop-down and give it a month at max. It works for a few days. I'll try to measure the time until it vanish. Experiment started right now.

![Screenshot from 2021-09-26 23-34-58](https://user-images.githubusercontent.com/2878126/134824934-2b00f3ff-5d8f-47d2-a641-43dff022a002.png)

northys:
@dhaavi hi there after 22 hours.

```

northys:
Interesting thing is that I have some domains allowed for a long time and I didn't have to allow them again. E.g. binance.com and the mentioned aws tracking code that broke aws console service search omnibox. analytics.google.com is gone though.

![Screenshot from 2021-09-29 15-57-23](https://user-images.githubusercontent.com/2878126/135283343-c11c1b8c-829b-4853-a145-9dd3205a14f8.png)

ppacher:
Thanks for you investigation on that topic @northys! I'll try to reproduce the issue but since it does not affect all rules it's still seems somewhat random. I'll keep you updated!

thegranddesign:
This effectively happens for me on all rules every time if there's something I can do to help debug this.  Obviously this bug makes portmaster effectively useless :joy:

[...]

PR #461, opened 2021-12-03
Improve profile getting and updating to fix interference

dhaavi:


profile/profile.go
```diff
@@ -140,6 +140,9 @@ type Profile struct { //nolint:maligned // not worth the effort
 	// Lifecycle Management
 	outdated   *abool.AtomicBool
 	lastActive *int64
+
+	// savedInternally is set to true for profiles that are saved internally.
+	savedInternally bool
```
ppacher, line 145: I feel like this is just a workaround for fixing an architectural issue with profiles and how they are saved. Otherwise, LGTM, nice we finally found and fixed that

ppacher (approved):

ppacher:
This PR will likely finally resolve https://github.com/safing/portmaster/issues/391

ppacher (approved):

ppacher (approved):

PR #476, opened 2021-12-16
Fix active profile invalidation and revision counting

dhaavi:


profile/profile-layered.go
```diff
@@ -165,6 +165,25 @@ func (lp *LayeredProfile) LocalProfile() *Profile {
 	return lp.localProfile
 }
 
+// RevisionCnt returns the current profile revision counter.
+func (lp *LayeredProfile) IncreaseRevisionCounter(lock bool) (revisionCounter uint64) {
+	if lp == nil {
+		return 0
```
ppacher, line 171: Comment is wrong and I'm not sure we should really export this method. Having a parameter `lock` that can be specified from outside the package seems error prone

ppacher (approved):

profile/profile-layered.go
dhaavi: Yup, will fix comment and make it internal.

ppacher (approved):

PR #492, opened 2022-01-12
Improve location estimation

dhaavi:


intel/geoip/fill_missing.go
```diff
@@ -1,5 +1,8 @@
 package geoip
 
+// ca ist the default country based accuracy radius.
+const ca = 200
+
```
ppacher, line 5: ^^ `ca` is IMHO not a meaningful name for this constant :D

intel/geoip/location.go
```diff
@@ -53,29 +60,30 @@ func (l *Location) EstimateNetworkProximity(to *Location) (proximity int) {
...
-		- Country match: 25
-		- ASOrg match: 20
-		- ASN match: 15
-		- Coordinate distance: 0-10
+		- Continent match: 25
+		- Country match: 20
+		- ASOrg match: 15
+		- ASN match: 10
+		- Coordinate distance: 0-30
 	*/
```
ppacher, line 68: just remove that comment and point to the `const` section above. This will for sure be wrong in a few weeks/months again

ppacher (changes requested):

ppacher (approved):

PR #514, opened 2022-02-02
Fix tests and linters

dhaavi:
- Update test script, config and workflow
- Fix tests and linters


dhaavi (commented):
Reviewed again. Looks good.

PR #535, opened 2022-02-17
Add SPN Settings

dhaavi:
- Add config option annotation key for verdict names
- Add endpoint list validation regex and function
- Add per app SPN settings


firewall/tunnel.go
```diff
@@ -1,23 +1,114 @@
...
+	// Tunnel all the things!
+
+	// Check if ready.
+	if !captain.ClientReady() {
+		// Block connection as SPN is not ready yet.
+		log.Tracer(pkt.Ctx()).Trace("SPN not ready for tunneling")
+		conn.Failed("SPN not ready for tunneling", "")
+		return
+	}
+
```
ppacher, line 90: shouldn't we just try to pause/wait until it is ready so apps don't get a bunch of failed-to-connect issues on startup?

firewall/tunnel.go
```diff
@@ -1,23 +1,114 @@
...
+	// Set options.
+	conn.TunnelOpts = &navigator.Options{
+		HubPolicies:                   layeredProfile.StackedExitHubPolicies(),
+		CheckHubExitPolicyWith:        conn.Entity,
+		RequireTrustedDestinationHubs: conn.Encrypted,
+		RoutingProfile:                layeredProfile.SPNRoutingAlgorithm(),
+	}
```
ppacher, line 97: I guess `RequireTrustedDestinationHubs` should be set to `!conn.Encrypted`?

firewall/tunnel.go
```diff
@@ -1,23 +1,114 @@
...
+	// Special handling for the internal DNS resolver.
+	if conn.Process().Pid == ownPID && resolver.IsResolverAddress(conn.Entity.IP, conn.Entity.Port) {
+		conn.TunnelOpts.RoutingProfile = navigator.RoutingProfileHomeID
+	}
+
+	// Queue request in sluice.
+	err := sluice.AwaitRequest(conn, crew.HandleSluiceRequest)
+	if err != nil {
+		log.Tracer(pkt.Ctx()).Warningf("failed to rqeuest tunneling: %s", err)
+		conn.Failed("failed to request tunneling", "")
```
ppacher, line 108: Typo in "request"

profile/config-update.go
```diff
@@ -75,6 +77,20 @@ func updateGlobalConfigProfile(ctx context.Context, task *modules.Task) error {
...
+		// TODO: module error?
+		lastErr = err
+	}
+
+	list = cfgOptionExitHubPolicy()
+	cfgSPNExitHubPolicy, err = endpoints.ParseEndpoints(list)
+	if err != nil {
+		// TODO: module error?
+		lastErr = err
+	}
```
ppacher, line 92: Consider using a multierr here so we don't miss errors and the user needs to try over and over again until all errors are fixed

profile/config.go
```diff
@@ -119,6 +137,39 @@ var securityLevelSettings = []string{
...
+		{Name: "Exclude Poland (PL)", Action: config.QuickMergeTop, Value: []string{"- PL"}},
+		{Name: "Exclude United Kingdom (GB)", Action: config.QuickMergeTop, Value: []string{"- GB"}},
+		{Name: "Exclude United States of America (US)", Action: config.QuickMergeTop, Value: []string{"- US"}},
+	}
+
+	// SPNRulesVerdictNames defines the verdicts names to be used for SPN Rules.
+	SPNRulesVerdictNames = map[string]string{
+		"-": "Exclude", // Default.
+		"+": "Allow",
+	}
```
ppacher, line 160: I feel like the verdicts are not very clear on what they mean. What happens in `Exclude`? Is the app excluded from using the SPN for that domain?

ppacher (changes requested):

PR #536, opened 2022-02-17
Update network rating setting wording

dhaavi:


status/config.go
```diff
@@ -10,9 +10,11 @@ var (
...
 	if err := config.Register(&config.Option{
-		Name:           "Enable Network Rating System",
-		Key:            CfgEnableNetworkRatingSystemKey,
-		Description:    "Enables the Network Rating System, which allows you to quickly increase security and privacy throughout the settings by changing your the network rating level in the top left. Please note that this feature is now in the sunset phase and will be replaced by a superior and easier to understand system in the future.",
+		Name: "Enable Network Rating System",
+		Key:  CfgEnableNetworkRatingSystemKey,
+		Description: `Enable the Network Rating System, which allows you to configure settings to be active in one environment but not in the other, like allowing sensitive connections at home but not at the public library.
+
+Please note that this feature is now in the sunset phase and will be replaced by a superior and easier to understand system in the future.`,
 		OptType:        config.OptTypeBool,
```
ppacher, line 18: I'm not sure people will understand what "sunset phase" means. I'd just call it "deprecated and will be removed"

ppacher (dismissed):

status/config.go
davegson: quick fix would be to remove the sunset phrasing, works too fmpov:

> Please note that this feature will be replaced by a superior ..

dhaavi: "sunset phase" wording was removed.

PR #547, opened 2022-02-25
Fixes and Improvements #14

dhaavi:
- Fix bypass detection to correctly attribute encrypted DNS bypassing
- Only hide successful queries of the system resolver
- Fix verdicts of DNS request connections
- Stop useless updating of app compat warning
- Add validation for DNS server config
- Enable build for arm64 platform

### DIFF
--- a/compat/notify.go
+++ b/compat/notify.go
@@ -148,14 +148,23 @@ func (issue *appIssue) notify(proc *process.Process) {
 
 	// Set warning on profile.
 	module.StartWorker("set app compat warning", func(ctx context.Context) error {
+		var changed bool
+
 		func() {
 			p.Lock()
 			defer p.Unlock()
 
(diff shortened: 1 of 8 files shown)


compat/notify.go
```diff
@@ -123,6 +124,13 @@ func (issue *appIssue) notify(proc *process.Process) {
 		proc.Path,
 	)
 
+	// Build message.
+	messageAppNameReplaces := make([]interface{}, strings.Count(issue.message, "%s"))
+	for i := range messageAppNameReplaces {
+		messageAppNameReplaces[i] = p.Name
+	}
+	message := fmt.Sprintf(issue.message, messageAppNameReplaces...)
+
```
ppacher, line 133: :D :+1:

ppacher (approved):

PR #572, opened 2022-03-16
Add in-memory SQLite database and basic query support

ppacher:


netquery/manager.go
```diff
@@ -0,0 +1,225 @@
...
+	extraData := map[string]interface{}{}
+
+	if conn.Entity != nil {
+		extraData["cname"] = conn.Entity.CNAME
+		extraData["blockedByLists"] = conn.Entity.BlockedByLists
+		extraData["blockedEntities"] = conn.Entity.BlockedEntities
+		extraData["reason"] = conn.Reason
+
+		c.RemoteIP = conn.Entity.IP.String()
+		c.RemotePort = conn.Entity.Port // FIXME(ppacher): or do we want DstPort() here?
```
dhaavi, line 198: conn.Entity.Port is correct.
DstPort() only exists for the rule engine.

netquery/manager.go
```diff
@@ -0,0 +1,225 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal extra data: %w", err)
+	}
+	c.ExtraData = extraDataBlob
+
+	return &c, nil
+}
+
+func genConnID(conn *network.Connection) string {
+	data := conn.ID + "-" + time.Unix(conn.Started, 0).String()
```
dhaavi, line 222: Should we use the date first, so it would automatically sort them chronologically should this data land in a NoSQL db at some point?
Also, I would use a proper format for the date string.

netquery/manager.go
```diff
@@ -0,0 +1,225 @@
...
+	// Though, that's still better than always pushing NEW for existing
+	// connections.
+	// If we would use UnixNano() here chances would be even worse.
+	//
+	// Verify if the check in portbase/api/database.go is vulnerable
+	// to such timing issues in general.
+	wrapper.SetMeta(&record.Meta{
+		Created:  conn.Started.Unix(),
+		Modified: time.Now().Unix(),
+	})
```
dhaavi, line 155: You can just set the data in the NewWrapper call above (and save one allocation).
Is there some way to get an indicator from SQLite whether the record was inserted or updated?

netquery/database.go
```diff
@@ -0,0 +1,298 @@
...
+	// the query API.
+	//
+	// Use ConvertConnection from this package to convert a network.Connection to this
+	// representation.
+	Conn struct {
+		// ID is a device-unique identifier for the connection. It is built
+		// from network.Connection by hashing the connection ID and the start
+		// time. We cannot just use the network.Connection.ID because it is only unique
+		// as long as the connection is still active and might be, although unlikely,
+		// reused afterwards.
```
dhaavi, line 63: I don't see any hashing? See comment below at getConnID().

netquery/module_api.go
```diff
@@ -0,0 +1,140 @@
...
+				Internal:  true,
+				CacheSize: 0,
+			})
+
+			sqlStore, err = NewInMemory()
+			if err != nil {
+				return fmt.Errorf("failed to create in-memory database: %w", err)
+			}
+
+			mng, err = NewManager(sqlStore, "netquery/updates/", runtime.DefaultRegistry)
```
dhaavi, line 41: Is "updates" good wording when this will also send out new connections?

netquery/module_api.go
```diff
@@ -0,0 +1,140 @@
...
+
+func init() {
+	var (
+		module   *modules.Module
+		db       *database.Interface
+		sqlStore *Database
+		mng      *Manager
+	)
+
+	module = modules.Register(
```
dhaavi, line 24: This function call is _huge_. Consider moving prep and start to separate functions. Also, does anything need to be stopped or unloaded when shutting down (or prob done by workers)?

netquery/module_api.go
```diff
@@ -0,0 +1,140 @@
...
+			}
+
+			return nil
+		},
+		/* Start Module */
+		func() error {
+			ch := make(chan *network.Connection, 100)
+
+			module.StartServiceWorker("netquery-feeder", time.Second, func(ctx context.Context) error {
+				sub, err := db.Subscribe(query.New("network:"))
```
dhaavi, line 53: The subscription service _may_ drop stuff if channels get full. Consider using a dedicated interface just for this.

netquery/module_api.go
```diff
@@ -0,0 +1,140 @@
...
+						}
+
+						ch <- conn
+					}
+				}
+			})
+
+			module.StartServiceWorker("netquery-persister", time.Second, func(ctx context.Context) error {
+				defer close(ch)
+
```
dhaavi, line 82: "ServiceWorker" means that it will be restart in case of an error. Closing the channel will kill this.

netquery/module_api.go
```diff
@@ -0,0 +1,140 @@
...
+				mng.HandleFeed(ctx, ch)
+				return nil
+			})
+
+			module.StartWorker("netquery-row-cleaner", func(ctx context.Context) error {
```
dhaavi, line 87: Consider using a ServiceWorker here.

netquery/module_api.go
```diff
@@ -0,0 +1,140 @@
...
+				for {
+					select {
+					case <-ctx.Done():
+						return nil
+					case <-time.After(10 * time.Second):
+						count, err := sqlStore.Cleanup(ctx, time.Now().Add(-5*time.Minute))
```
dhaavi, line 93: The current time windows is 10 minutes - can you get that as a constant from the network package?

netquery/module_api.go
```diff
@@ -0,0 +1,140 @@
...
+					}
+				}
+			})
+
+			// for debugging, we provide a simple direct SQL query interface using
+			// the runtime database
+			_, err := NewRuntimeQueryRunner(sqlStore, "netquery/query/", runtime.DefaultRegistry)
+			if err != nil {
+				return fmt.Errorf("failed to set up runtime SQL query runner: %w", err)
+			}
```
dhaavi, line 130: Do you maybe want to put these behind a flag - so we can keep and them and don't forget removing them.

netquery/module_api.go
```diff
@@ -0,0 +1,140 @@
...
+
+			return nil
+		},
+		nil,
+		"network",
+		"database",
+	)
+
+	module.Enable()
```
dhaavi, line 139: Not sure why this is here. Let's discuss.

netquery/runtime_query_runner.go
```diff
@@ -0,0 +1,77 @@
...
+// the record key.
+type RuntimeQueryRunner struct {
+	db        *Database
+	reg       *runtime.Registry
+	keyPrefix string
+}
+
+// NewRuntimeQueryRunner returns a new runtime SQL query runner that parses
+// and serves SQL queries form GET <prefix>/<plain sql query> requests.
+func NewRuntimeQueryRunner(db *Database, prefix string, reg *runtime.Registry) (*RuntimeQueryRunner, error) {
```
dhaavi, line 33: What is the benefit of handling this via the db api and not the http api?

dhaavi (changes requested):

netquery/manager.go
dhaavi: Copy meta from connection.

netquery/database.go
dhaavi: My bad.

dhaavi (approved):

netquery/module_api.go
```diff
@@ -0,0 +1,187 @@
...
+
+	if err := api.RegisterEndpoint(api.Endpoint{
+		Path:        "netquery/charts/connection-active",
+		MimeType:    "application/json",
+		Read:        api.PermitAnyone,
+		Write:       api.PermitAnyone,
+		BelongsTo:   m.Module,
+		HandlerFunc: chartHandler.ServeHTTP,
+		Name:        "Query In-Memory Database",
+		Description: "Query the in-memory sqlite database",
```
dhaavi, line 92: Update name and description.

netquery/module_api.go
```diff
@@ -0,0 +1,187 @@
...
+	chartHandler := &ChartHandler{
+		Database: m.sqlStore,
+	}
+
+	// FIXME(ppacher): use appropriate permissions for this
+	if err := api.RegisterEndpoint(api.Endpoint{
+		Path:        "netquery/query",
+		MimeType:    "application/json",
+		Read:        api.PermitAnyone,
+		Write:       api.PermitAnyone,
```
dhaavi, line 75: Constrain to at least user, as Anyone makes this publicly available.

netquery/module_api.go
```diff
@@ -0,0 +1,187 @@
...
+		Description: "Query the in-memory sqlite database",
+	}); err != nil {
+		return fmt.Errorf("failed to register API endpoint: %w", err)
+	}
+
+	if err := api.RegisterEndpoint(api.Endpoint{
+		Path:        "netquery/charts/connection-active",
+		MimeType:    "application/json",
+		Read:        api.PermitAnyone,
+		Write:       api.PermitAnyone,
```
dhaavi, line 88: Constrain to at least user, as Anyone makes this publicly available.

netquery/module_api.go
```diff
@@ -0,0 +1,187 @@
...
+		for {
+			select {
+			case <-ctx.Done():
+				return nil
+			case <-time.After(1 * time.Second):
+				count, err := mod.sqlStore.CountRows(ctx)
+				if err != nil {
+					log.Errorf("netquery: failed to count number of rows in memory: %s", err)
+				} else {
+					log.Infof("netquery: currently holding %d rows in memory", count)
```
dhaavi, line 161: Maybe move to debug log level.

netquery/module_api.go
```diff
@@ -0,0 +1,187 @@
...
+				if err != nil {
+					log.Errorf("netquery: failed to count number of rows in memory: %s", err)
+				} else {
+					log.Infof("netquery: successfully removed %d old rows that ended before %s", count, threshold)
+				}
+			}
+		}
+	})
+
+	mod.StartWorker("netquery-row-counter", func(ctx context.Context) error {
```
dhaavi, line 151: Should we only start this in certain cases? No need for this in production.

netquery/module_api.go
```diff
@@ -0,0 +1,187 @@
...
+			select {
+			case <-ctx.Done():
+				return nil
+			case <-time.After(10 * time.Second):
+				threshold := time.Now().Add(-network.DeleteConnsAfterEndedThreshold)
+				count, err := mod.sqlStore.Cleanup(ctx, threshold)
+				if err != nil {
+					log.Errorf("netquery: failed to count number of rows in memory: %s", err)
+				} else {
+					log.Infof("netquery: successfully removed %d old rows that ended before %s", count, threshold)
```
dhaavi, line 145: Move to trace?

netquery/module_api.go
```diff
@@ -0,0 +1,187 @@
...
+	}); err != nil {
+		return fmt.Errorf("failed to register API endpoint: %w", err)
+	}
+
+	return nil
+}
+
+func (mod *Module) Start() error {
+	mod.StartServiceWorker("netquery-feeder", time.Second, func(ctx context.Context) error {
+		sub, err := mod.db.Subscribe(query.New("network:"))
```
dhaavi, line 102: Is there a better option than subscribing via the database?
Especially when we want to get rid of that interface.

netquery/module_api.go
```diff
@@ -0,0 +1,187 @@
...
+						log.Errorf("netquery: failed to dump sqlite memory content: %w", err)
+					}
+				*/
+			}
+		}
+	})
+
+	// for debugging, we provide a simple direct SQL query interface using
+	// the runtime database
+	_, err := NewRuntimeQueryRunner(mod.sqlStore, "netquery/query/", runtime.DefaultRegistry)
```
dhaavi, line 175: Is this is a security risk? Should we deactivate in production?

netquery/query.go
```diff
@@ -0,0 +1,556 @@
...
+		Page     int `json:"page"`
+	}
+)
+
+func (query *Query) UnmarshalJSON(blob []byte) error {
+	if *query == nil {
+		*query = make(Query)
+	}
+
+	var model map[string]json.RawMessage
```
dhaavi, line 99: Discuss raw json parsing.

netquery/query.go
```diff
@@ -0,0 +1,556 @@
...
+		return err
+	}
+
+	for columnName, rawColumnQuery := range model {
+		if len(rawColumnQuery) == 0 {
+			continue
+		}
+
+		switch rawColumnQuery[0] {
+		case '{':
```
dhaavi, line 111: Does this require a json formatted input without spacing?

dhaavi (changes requested):
Very nice work!
Some minor comments and questions.

netquery/manager.go
```diff
@@ -0,0 +1,263 @@
...
+// convertConnection converts conn to the local representation used
+// to persist the information in SQLite. convertConnection attempts
+// to lock conn and may thus block for some time.
+func convertConnection(conn *network.Connection) (*Conn, error) {
+	conn.Lock()
+	defer conn.Unlock()
+	direction := "outbound"
+	if conn.Inbound {
+		direction = "inbound"
+	}
```
dhaavi, line 161: This quite a bit for a boolean. Should we simplify this?

PR #575, opened 2022-03-19
Improve debug data

dhaavi:
- Add debug interfaces for wfp and iptables
- Mark config options as sensitive


dhaavi:
Requires https://github.com/safing/portbase/pull/157 and https://github.com/safing/spn/pull/63.

ppacher (dismissed):

PR #586, opened 2022-03-30
Reduce compat false positivies

dhaavi:
- Increase self-check timeouts and threshold
- Fix secure DNS bypass notification to reflect reality
- Add notify threshold for app related notification
- Fix connection locking in firewall packet handler

### DIFF
--- a/network/connection.go
+++ b/network/connection.go
@@ -580,10 +580,7 @@ func (conn *Connection) SetFirewallHandler(handler FirewallHandler) {
 		conn.pktQueue = make(chan packet.Packet, 1000)
 
 		// start handling
-		module.StartWorker("packet handler", func(ctx context.Context) error {
-			conn.packetHandler()
-			return nil
-		})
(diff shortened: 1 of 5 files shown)


network/connection.go
```diff
@@ -614,29 +614,30 @@ func (conn *Connection) packetHandler() {
...
 
-		// execute handler or verdict
-		if conn.firewallHandler != nil {
-			conn.firewallHandler(conn, pkt)
-		} else {
-			defaultFirewallHandler(conn, pkt)
-		}
+		func() {
+			conn.Lock()
+			defer conn.Unlock()
```
ppacher, line 621: Performance wise I would not use an anonymous function here in the hot-path but rather use a normal, named function

ppacher (changes requested):

dhaavi:
Linter failed with internal error.

ppacher (dismissed):

ppacher (approved):

PR #585, opened 2022-03-30
Clear DNS cache when DNS resolver config changes and update resolver config

dhaavi:
- Clear DNS cache when DNS resolver config changes in any way
- Improve resolver config for intro slides

PR #590, opened 2022-04-04
Change order of DNS server quick-settings and enable second cloudflare resolver by default

ppacher:
Together with safing/portmaster-ui#219


ppacher:
Changes are already part of #585 .

PR #599, opened 2022-04-13
DoH support

vlabo:
There are still some improvements to be made but it's a working version from my tests.

I would be nice if there is any comments or suggestions and will you be willing to merge this feature to portmaster?

Planned improvements:  
* Keep the https connection open so we don't need to create a new one for each query
* Better user interface for the dns settings
     * enter a domain name (instead of ip) for DoT and DoH and fetch the ips before starting portmaster (maybe this needs to be separate PR)
     * Some DoH providers don't fallow [the standard "dns-query" url path](https://www.rfc-editor.org/rfc/rfc8484.html)  like applied-privacy.net. There needs to be a way of specifying the custom path


dhaavi:
Hey @vlabo, thanks for the PR - this is great! We've been wanting to support this for some time now, but the resources didn't allow it.

What do you think about going over this PR when you come to the office? I think that would be a great task to work on together - as you don't have to immerse yourself into a new topic.

I'll just jot down some ideas and topics here, so I don't forget them - we'll then go through them together:

- Use path from URL
- Integrate config into http.Client (as we don't handle the connection ourselves)
- Minimize http request metadata
- Reuse connections via http.Client
  - Can we log when connections are created/dropped as with DoT?
- Add "https" alt so we can just paste DoH URLs as is
- Think about how to best "pre resolve"
  - On startup?
  - Bypass resolves that are not ready?
- HTTP/3?

resolver/resolver-https.go
```diff
@@ -0,0 +1,116 @@
...
+	request := &http.Request{
+		Method:     "GET",
+		URL:        url,
+		Proto:      "HTTP/1.1",
+		ProtoMajor: 1,
+		ProtoMinor: 1,
+		Header:     make(http.Header),
+		Body:       nil,
+		Host:       hr.resolver.ServerAddress,
+	}
```
ppacher, line 89: Use `http.NewRequestWithContext` instead

its0x08 (approved):

resolver/resolvers.go
```diff
@@ -200,10 +212,162 @@ func createResolver(resolverURL, source string) (*Resolver, bool, error) {
...
+
+	if scheme == ServerTypeDoT || scheme == ServerTypeDoH {
+
+		switch {
+		case isHostnameDomain && verifyDomain != "":
+			return fmt.Errorf("cannot have verify parameter with a domain as a hostname")
+		case !isHostnameDomain && verifyDomain == "":
+			return fmt.Errorf("verify paremeter must be set when using ip as domain")
+		case !isHostnameDomain && paramterServerIP != "":
+			return fmt.Errorf("cannot have ip parameter while domain is an ip")
```
dhaavi, line 253: suggest: `cannot set the IP address via both the hostname in the URL and the ip parameter`

resolver/resolvers.go
```diff
@@ -200,10 +212,162 @@ func createResolver(resolverURL, source string) (*Resolver, bool, error) {
...
+		}
+	} else {
+		if verifyDomain != "" {
+			return fmt.Errorf("domain verification only supported in DoT and DoH")
```
dhaavi, line 257: suggest: `domain verification is only supported by DoT and DoH servers`

resolver/resolvers.go
```diff
@@ -200,10 +212,162 @@ func createResolver(resolverURL, source string) (*Resolver, bool, error) {
...
+			return fmt.Errorf("DoT must have a verify query parameter set")
+		}
+	}
+
+	if scheme != ServerTypeDoH {
+		path := query.Get(parameterPath)
+		if path != "" {
+			return fmt.Errorf("path parameter is only supported in DoH")
+		}
+	}
```
dhaavi, line 269: L264-269: Use the path of the URL instead of an additional parameter.

resolver/resolvers.go
```diff
@@ -200,10 +212,162 @@ func createResolver(resolverURL, source string) (*Resolver, bool, error) {
...
+		case !isHostnameDomain && paramterServerIP != "":
+			return fmt.Errorf("cannot have ip parameter while domain is an ip")
+		}
+	} else {
+		if verifyDomain != "" {
+			return fmt.Errorf("domain verification only supported in DoT and DoH")
+		}
+		if verifyDomain == "" && !isHostnameDomain {
+			return fmt.Errorf("DoT must have a verify query parameter set")
+		}
```
dhaavi, line 261: What are you testing here?
DoT will not go down this path, you are in the "else" clause.

resolver/resolvers.go
```diff
@@ -92,65 +98,60 @@ func createResolver(resolverURL, source string) (*Resolver, bool, error) {
...
+			serverAddress = net.JoinHostPort(paramterServerIP, strconv.Itoa(int(port)))
+			verifyDomain = u.Hostname()
+		case !isHostnameDomain && verifyDomain != "": // ip and domain as parameter
+			serverAddress = net.JoinHostPort(ip.String(), strconv.Itoa(int(port)))
+		case isHostnameDomain && verifyDomain == "" && paramterServerIP == "": // only domain
+			verifyDomain = u.Hostname()
+		}
+	} else {
+		serverAddress = net.JoinHostPort(ip.String(), strconv.Itoa(int(port)))
+	}
```
dhaavi, line 145: L132-145: Would it make sense to merge this with `checkURLParameterValidity` and have all options, parsing and validation in there?

resolver/resolvers.go
```diff
@@ -92,65 +98,60 @@ func createResolver(resolverURL, source string) (*Resolver, bool, error) {
...
-			return nil, false, fmt.Errorf(`unknown parameter "%s"`, key)
-		}
+	err = checkURLParameterValidity(u.Scheme, isHostnameDomain, query)
+	if err != nil {
+		return nil, false, err
 	}
 
 	// Check domain verification config.
+	serverAddress := ""
+	paramterServerIP := query.Get(parameterIP)
```
dhaavi, line 129: typo: `parameterServerIP`
suggest: `serverIPParameter` as to not confuse with the `parameter*` constants.

Also, the IP address should be validated.

resolver/resolvers.go
```diff
@@ -92,65 +98,60 @@ func createResolver(resolverURL, source string) (*Resolver, bool, error) {
...
+		}
+	} else {
+		serverAddress = net.JoinHostPort(ip.String(), strconv.Itoa(int(port)))
+	}
+
+	// Check path for https (doh) request
+	path := query.Get(parameterPath)
+	if u.Path != "" {
+		path = u.Path
 	}
```
dhaavi, line 151: Use `u.Path` only.

resolver/resolvers.go
```diff
@@ -92,65 +98,60 @@ func createResolver(resolverURL, source string) (*Resolver, bool, error) {
...
-	if verifyDomain == "" && u.Scheme == ServerTypeDoT {
-		return nil, false, fmt.Errorf("DOT must have a verify query parameter set")
+
+	if path != "" && !strings.HasPrefix(path, "/") {
+		path = "/" + path
 	}
```
dhaavi, line 155: Why is this check necessary?

resolver/resolvers.go
```diff
@@ -200,10 +212,162 @@ func createResolver(resolverURL, source string) (*Resolver, bool, error) {
...
+}
+
+func parsePortFromURL(url *url.URL) (uint16, error) {
+	var port uint16
+	hostPort := url.Port()
+	if hostPort != "" {
+		// There is a port in the url
+		parsedPort, err := strconv.ParseUint(hostPort, 10, 16)
+		if err != nil {
+			return 0, fmt.Errorf("resolver port %q invalid", url.Port())
```
dhaavi, line 351: suggest: `invalid port %q`

resolver/resolvers.go
```diff
@@ -200,10 +212,162 @@ func createResolver(resolverURL, source string) (*Resolver, bool, error) {
...
+			return fmt.Errorf("path parameter is only supported in DoH")
+		}
+	}
+
+	return nil
+}
+
+func checkAndResolveServerAddressAndPort(resolver *Resolver) error {
+	if resolver.ServerAddress == "" {
+		resolverIps, err := resolveDomainIP(context.Background(), resolver.VerifyDomain)
```
dhaavi, line 276: Instead of resolving here, just return `ErrContinue` for that domain in `Query`, then the next resolver will be used for that query and all the caching will be used too.

dhaavi (changes requested):

resolver/resolver-https.go
```diff
@@ -0,0 +1,140 @@
...
+	newResolver.BasicResolverConn.init()
+	return newResolver
+}
+
+// Query executes the given query against the resolver.
+func (hr *HttpsResolver) Query(ctx context.Context, q *Query) (*RRCache, error) {
+
+	// Do not resolve domain names that are needed to initialize a resolver
+	if hr.resolver.Info.IP == nil {
+		if _, ok := resolverInitDomains[q.FQDN[:len(q.FQDN)-1]]; ok {
```
dhaavi, line 73: Instead of tricking around here, please save the fully qualified domain name (ie. including the trailing `.`) in `resolverInitDomains`.
You can use `dns.Fqdn` for that.

resolver/resolver-https.go
```diff
@@ -0,0 +1,140 @@
...
+	if err != nil {
+		return nil, err
+	}
+	b64dns := base64.RawStdEncoding.EncodeToString(buf)
+
+	// Set the host, if we dont have IP address just use the domain
+	host := hr.resolver.ServerAddress
+	if host == "" {
+		host = net.JoinHostPort(hr.resolver.VerifyDomain, strconv.Itoa(int(hr.resolver.Info.Port)))
+	}
```
dhaavi, line 92: Should not `resolver.ServerAddress` already hold the ready to use version?

resolver/resolvers.go
```diff
@@ -28,18 +28,21 @@ type Scope struct {
...
-	systemResolvers []*Resolver          // all resolvers that were assigned by the system
-	localScopes     []*Scope             // list of scopes with a list of local resolvers that can resolve the scope
-	activeResolvers map[string]*Resolver // lookup map of all resolvers
-	resolversLock   sync.RWMutex
+	globalResolvers     []*Resolver          // all (global) resolvers
+	localResolvers      []*Resolver          // all resolvers that are in site-local or link-local IP ranges
+	systemResolvers     []*Resolver          // all resolvers that were assigned by the system
+	localScopes         []*Scope             // list of scopes with a list of local resolvers that can resolve the scope
+	activeResolvers     map[string]*Resolver // lookup map of all resolvers
+	resolverInitDomains map[string]bool      // a set with all domains of the dns resolvers
```
dhaavi, line 44: Use `map[string]struct{}` instead.
See eg. https://medium.com/easyread/golang-series-empty-struct-ed317e6d8600

resolver/resolvers.go
```diff
@@ -91,93 +96,63 @@ func createResolver(resolverURL, source string) (*Resolver, bool, error) {
...
 
+	if resolverInitDomains == nil {
+		resolverInitDomains = make(map[string]bool)
+	}
+
 	switch u.Scheme {
-	case ServerTypeDNS, ServerTypeDoT, ServerTypeTCP:
+	case ServerTypeDNS, ServerTypeDoT, ServerTypeDoH, ServerTypeTCP:
+	case HttpsProtocol:
+		u.Scheme = ServerTypeDoH
```
dhaavi, line 106: Please also create an alias for `tls` -> `dot` to support the notation that eg. Quad9 uses.

resolver/resolvers.go
```diff
@@ -204,6 +179,106 @@ func createResolver(resolverURL, source string) (*Resolver, bool, error) {
 	return newResolver, false, nil
 }
 
+func checkAndSetResolverParamters(u *url.URL, resolver *Resolver) error {
+
+	// Check if we are using domain name and if it's in a valid scheme
+	ip := net.ParseIP(u.Hostname())
+	hostnameIsDomaion := (ip == nil)
+	if ip == nil && u.Scheme != ServerTypeDoH && u.Scheme != ServerTypeDoT {
+		return fmt.Errorf("resolver IP %q invalid", u.Hostname())
```
dhaavi, line 188: suggest: "is invalid"
Full sentences are easier to read and understand.

resolver/resolvers.go
```diff
@@ -204,6 +179,106 @@ func createResolver(resolverURL, source string) (*Resolver, bool, error) {
...
+			parameterVerify,
+			parameterIP,
+			parameterBlockedIf,
+			parameterSearch,
+			parameterSearchOnly,
+			parameterPath:
+			// Known key, continue.
+		default:
+			// Unknown key, abort.
+			return fmt.Errorf(`unknown parameter "%s"`, key)
```
dhaavi, line 212: Probably was me, but `%q` is better than `"%s"`, as it escapes values inside.

resolver/resolvers.go
```diff
@@ -204,6 +179,106 @@ func createResolver(resolverURL, source string) (*Resolver, bool, error) {
...
+			resolver.Info.IP = net.ParseIP(paramterServerIP)
+			resolver.ServerAddress = net.JoinHostPort(paramterServerIP, strconv.Itoa(int(resolver.Info.Port)))
+			resolver.VerifyDomain = u.Hostname()
+		case !hostnameIsDomaion && resolver.VerifyDomain != "": // ip and domain as parameter
+			resolver.ServerAddress = net.JoinHostPort(ip.String(), strconv.Itoa(int(resolver.Info.Port)))
+		case hostnameIsDomaion && resolver.VerifyDomain == "" && paramterServerIP == "": // only domain
+			resolver.VerifyDomain = u.Hostname()
+		}
+
+		resolver.Info.Domain = resolver.VerifyDomain
```
dhaavi, line 243: Do we need both `Domain` and `VerifyDomain`? If not, please go for one of them, you decide.
I think we always use the domain for verification if we have one.

resolver/resolvers.go
```diff
@@ -204,6 +179,106 @@ func createResolver(resolverURL, source string) (*Resolver, bool, error) {
...
+		// Parse and set IP and Domain to the resolver
+		switch {
+		case hostnameIsDomaion && paramterServerIP != "": // domain and ip as parameter
+			resolver.Info.IP = net.ParseIP(paramterServerIP)
+			resolver.ServerAddress = net.JoinHostPort(paramterServerIP, strconv.Itoa(int(resolver.Info.Port)))
+			resolver.VerifyDomain = u.Hostname()
+		case !hostnameIsDomaion && resolver.VerifyDomain != "": // ip and domain as parameter
+			resolver.ServerAddress = net.JoinHostPort(ip.String(), strconv.Itoa(int(resolver.Info.Port)))
+		case hostnameIsDomaion && resolver.VerifyDomain == "" && paramterServerIP == "": // only domain
+			resolver.VerifyDomain = u.Hostname()
```
dhaavi, line 240: You could set `ServerAddress` here with the domain name, then you don't have check again later.

dhaavi (changes requested):
Some minor comments.

I'd also like to go through the code and look at style together.

resolver/resolvers.go
vlabo: The reason I added Domain in Info is for the `func (info *ResolverInfo) ID()` function
So we can have a proper info in the additional section

;; ADDITIONAL SECTION:
info.portmaster.        0       IN      TXT     "accepted: allowing dns request"
info.portmaster.        0       IN      TXT     "freshly resolved by Cloudflare (dot://cloudflare-dns.com:853#config)"

We also have a duplicate for the IP address in there. Im not sure what was the original idea of the separate Info struct and do we really need it

resolver/resolver-https.go
```diff
@@ -0,0 +1,132 @@
...
+		Extra:    reply.Extra,
+		Resolver: resolverInfo.Copy(),
+	}
+}
+
+// NewHTTPSResolver returns a new HttpsResolver.
+func NewHTTPSResolver(resolver *Resolver) *HttpsResolver {
+	tr := &http.Transport{}
+
+	if resolver.Info.IP != nil {
```
dhaavi, line 44: What are you trying to achieve with this check?

resolver/resolver.go
```diff
@@ -30,6 +30,11 @@ const (
 	ServerSourceEnv             = "env"
 )
 
+const (
+	HttpsProtocol = "https"
+	TlsProtocol   = "tls"
```
dhaavi, line 35: `TLS` should be capitalized. Normally the linter should catch this.

resolver/resolver.go
```diff
@@ -30,6 +30,11 @@ const (
 	ServerSourceEnv             = "env"
 )
 
+const (
+	HttpsProtocol = "https"
```
dhaavi, line 34: `HTTP` should be capitalized. Normally the linter should catch this.

resolver/resolver.go
```diff
@@ -135,6 +157,12 @@ func (info *ResolverInfo) DescriptiveName() string {
 			info.Name,
 			info.ID(),
 		)
+	case info.IP == nil:
```
dhaavi, line 160: Would recommend to check for `info.Domain != ""`, as this would prioritize the domain name if both Domain and IP are present.

dhaavi (changes requested):

resolver/resolvers.go
dhaavi: I see.
Then we can just drop `VerifyDomain` and start using `info.Domain`, right?

I could not find where we duplicate the IP address. Can you point me to it?

dhaavi (commented):
Changes look good so far.

resolver/resolvers.go
```diff
@@ -204,6 +181,107 @@ func createResolver(resolverURL, source string) (*Resolver, bool, error) {
...
+			resolver.ServerAddress = net.JoinHostPort(paramterServerIP, strconv.Itoa(int(resolver.Info.Port)))
+			resolver.Info.Domain = u.Hostname()
+		case !hostnameIsDomaion && resolver.Info.Domain != "": // ip and domain as parameter
+			resolver.ServerAddress = net.JoinHostPort(ip.String(), strconv.Itoa(int(resolver.Info.Port)))
+		case hostnameIsDomaion && resolver.Info.Domain == "" && paramterServerIP == "": // only domain
+			resolver.Info.Domain = u.Hostname()
+			resolver.ServerAddress = net.JoinHostPort(resolver.Info.Domain, strconv.Itoa(int(port)))
+		}
+
+		resolverInitDomains[dns.Fqdn(resolver.Info.Domain)] = struct{}{}
```
dhaavi, line 247: I think we should only set this when there is no IP.

dhaavi (changes requested):
Almost there!

PR #602, opened 2022-04-15
Add option to disable seamless DNS integration

dhaavi:
- Add config option to disable dns query interception
- Block DNS requests if bypass prevention is active
- Show notification about manual DNS setup instead of compatibility notice
- Use more verbose names for iptables chains


ppacher (dismissed):
I'd remove the C-17* stuff completely and rather give them descriptive names. Otherwise LGTM

PR #662, opened 2022-05-20
QOL Fixes in compat, netenv, dns and broadcasting

dhaavi:
- Reset system self-check after network change
- Calm down and relax dns query check


dhaavi:
Failing linter error is due to linter version mismatch, will fix later.

PR #751, opened 2022-07-19
Add endpoint to retrieve the current profile ID

ppacher:


core/api.go
```diff
@@ -55,6 +67,48 @@ func registerAPIEndpoints() error {
...
+		Read:       api.PermitAnyone,
+		BelongsTo:  module,
+		StructFunc: authorizeApp,
+		Name:       "Call to request an authentication token",
+		Parameters: []api.Parameter{
+			{
+				Method:      http.MethodGet,
+				Field:       "app-name",
+				Description: "The name of the application requesting access",
+			},
```
dhaavi, line 81: Should we instead derive the name from the requesting process path? Just supplying a name makes it really easy to fool the user.

core/api.go
```diff
@@ -99,3 +153,160 @@ func debugInfo(ar *api.Request) (data []byte, err error) {
...
+		return nil, 0, errors.New("invalid IP address")
+	}
+
+	port, err := strconv.ParseUint(portString, 10, 16)
+	if err != nil {
+		return nil, 0, err
+	}
+
+	return ip, uint16(port), nil
+}
```
dhaavi, line 213: This looks like a good candidate for our `netutils` package.

core/api.go
```diff
@@ -99,3 +153,160 @@ func debugInfo(ar *api.Request) (data []byte, err error) {
...
+		Version:  packet.IPv4,
+		Protocol: packet.TCP,
+		Src:      remoteIP,   // source as in the process we are looking for
+		SrcPort:  remotePort, // source as in the process we are looking for
+	}
+
+	proc, _, err := process.GetProcessByConnection(ar.Context(), pkt)
+	if err != nil {
+		return nil, err
+	}
```
dhaavi, line 186: L168-186: This looks like it belongs into the `process` package.

core/api.go
```diff
@@ -99,3 +153,160 @@ func debugInfo(ar *api.Request) (data []byte, err error) {
...
+func getPermission(p string) api.Permission {
+	switch p {
+	case "user":
+		return api.PermitUser
+	case "admin":
+		return api.PermitAdmin
+	default:
+		return api.NotSupported
+	}
+}
```
dhaavi, line 166: This looks like it belongs in the `api` package.

core/api.go
```diff
@@ -99,3 +153,160 @@ func debugInfo(ar *api.Request) (data []byte, err error) {
...
+			return nil, err
+		}
+	}
+
+	if getPermission(readPermStr) <= api.NotSupported {
+		return nil, fmt.Errorf("invalid read permission")
+	}
+	if getPermission(writePermStr) <= api.NotSupported {
+		return nil, fmt.Errorf("invalid read permission")
+	}
```
dhaavi, line 234: L229-234: This depends on `getPermission` not returning `PermitSelf`. This should be checked here instead.

core/api.go
```diff
@@ -99,3 +153,160 @@ func debugInfo(ar *api.Request) (data []byte, err error) {
...
+
+	// appIcon := ar.Request.URL.Query().Get("app-icon")
+	// TODO(ppacher): get the guessed mime-type from appIcon and make sure it's an image
```
dhaavi, line 237: I suggest to get the process to display name (and later) icon.

dhaavi (changes requested):
Code looks good, some minor comments.

Structure could be improved, as well as documentation. ;)

dhaavi (approved):

Issue #495, opened 2022-01-18
wayland support, either pass electron flags or add a `--use-wayland` flag to app args


skewballfox:
normally you can run any v12+ electron app by appending the flags `--enable-features=UseOzonePlatform --ozone-platform=wayland` to whatever command is launching the app. often these aren't really handled by the application itself but passed directly to electron (vscode actually complains at startup that these options aren't recognized)

it seems that `portmaster-start app`  isn't launching directly as an electron app, because adding these flags cause the command to fail. If you would rather avoid just passing random flags to the underlying engine, you could add a `--use-wayland` flag to append these specific arguments to the process that is starting electron

**Pre-Submit Checklist**:

- Check applicable sources for existing issues:
  - [Linux Compatibility](https://docs.safing.io/portmaster/install/linux#compatibility)
  - [VPN Compatibility](https://docs.safing.io/portmaster/install/status/vpn-compatibility)
  - [Github Issues](https://github.com/safing/portmaster/issues?q=is%3Aissue+label%3Ain%2Fcompatibility)

**What worked?**

launching the app via `/opt/safing/portmaster/portmaster-start app --data=/opt/safing/portmaster`

**What did not work?**

launching the app via `/opt/safing/portmaster/portmaster-start app --data=/opt/safing/portmaster --enable-features=UseOzonePlatform --ozone-platform=wayland`

**Debug Information**:

portmaster version: 0.7

ppacher:
hi @skewballfox, thanks for the report!

You should be able to pass any options/arguments to electron by using a `--` delimiter. After that separator all remaining arguments should be passed to electron. 

Like:

```
portmaster-start app --data=/path/to/data -- arg1ForElectron arg2ForElectron 
```

So maybe try the following:

```
/opt/safing/portmaster/portmaster-start app --data=/opt/safing/portmaster -- --enable-features=UseOzonePlatform --ozone-platform=wayland
```

skewballfox:
thank you, tested it out last night, opens in a native wayland window as expected.

PR #752, opened 2022-07-19
Automatically append arguments required by Electron to run on wayland

ppacher:
This should fix https://github.com/safing/portmaster-ui/issues/265 and is related to https://github.com/safing/portmaster/issues/495

### DIFF
--- a/cmds/portmaster-start/run.go
+++ b/cmds/portmaster-start/run.go
@@ -122,17 +122,18 @@ func getExecArgs(opts *Options, cmdArgs []string) []string {
 		args = append(args, "--input-signals")
 	}
 
-	// see https://www.freedesktop.org/software/systemd/man/pam_systemd.html#type=
-	if xdgSessionType := os.Getenv("XDG_SESSION_TYPE"); xdgSessionType == "wayland" && opts.Identifier == "app/portmaster-app.zip" {
-		// we're running the Portmaster UI App under Wayland so make sure we add some arguments
-		// required by Electron
-		args = append(args,
-			[]string{
-				"--enable-features=UseOzonePlatform",
-				"--ozone-platform=wayland",
-			}...,
-		)
-
+	if opts.Identifier == "app/portmaster-app.zip" {
+		// see https://www.freedesktop.org/software/systemd/man/pam_systemd.html#type=
+		if xdgSessionType := os.Getenv("XDG_SESSION_TYPE"); xdgSessionType == "wayland" {
+			// we're running the Portmaster UI App under Wayland so make sure we add some arguments
+			// required by Electron
+			args = append(args,
+				[]string{
+					"--enable-features=UseOzonePlatform",
+					"--ozone-platform=wayland",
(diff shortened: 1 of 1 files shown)


cmds/portmaster-start/run.go
```diff
@@ -121,6 +121,20 @@ func getExecArgs(opts *Options, cmdArgs []string) []string {
 	if stdinSignals {
 		args = append(args, "--input-signals")
 	}
+
+	// see https://www.freedesktop.org/software/systemd/man/pam_systemd.html#type=
+	if xdgSessionType := os.Getenv("XDG_SESSION_TYPE"); xdgSessionType == "wayland" && opts.Identifier == "app/portmaster-app.zip" {
```
dhaavi, line 126: Not sure about the impact of `os.Getenv`, but we could pull the `opts.Identifier == "app/portmaster-app.zip"` check outside and only check the ENV if we are actually trying to launch the app.

cmds/portmaster-start/run.go
```diff
@@ -121,6 +121,20 @@ func getExecArgs(opts *Options, cmdArgs []string) []string {
...
+		// we're running the Portmaster UI App under Wayland so make sure we add some arguments
+		// required by Electron
+		args = append(args,
+			[]string{
+				"--enable-features=UseOzonePlatform",
+				"--ozone-platform=wayland",
+			}...,
+		)
+
```
dhaavi, line 135: No idea what's up here, but the linter in the editor should show.

cmds/portmaster-start/run.go
```diff
@@ -121,6 +121,20 @@ func getExecArgs(opts *Options, cmdArgs []string) []string {
 	if stdinSignals {
 		args = append(args, "--input-signals")
 	}
+
+	// see https://www.freedesktop.org/software/systemd/man/pam_systemd.html#type=
```
dhaavi, line 125: Does that mean this this only works when systemd is used?

dhaavi (changes requested):

cmds/portmaster-start/run.go
ppacher: that environment variable is populated by systemd. I did not find any other reasonable way of detecting if wayland is used (and not just available or running for some other session). Most systems are using systemd anyway so this fix is for them. Non-systemd users can still workaround the issue by appending the args manually as suggested in  https://github.com/safing/portmaster/issues/495

cmds/portmaster-start/run.go
ppacher: environment variables are already part of the process memory and it's basically just a map lookup under a RLock/using a syscall so this check is negligible.

cmds/portmaster-start/run.go
ppacher: Though, for example GNOME also depends on that environment variable to decide in which mode to run so relying on that should be reasonable

dhaavi (approved):

PR #773, opened 2022-08-02
Add connection pooling for sqlite database access

ppacher:


netquery/database.go
```diff
@@ -48,6 +49,9 @@ type (
 		Schema *orm.TableSchema
 
 		pool *puddle.Pool[*sqlite.Conn]
```
dhaavi, line 51: Would suggest something like `readConnPool` in order to be more explicit.

dhaavi (approved):

dhaavi (approved):

PR #776, opened 2022-08-02
Improve custom filter list config key

dhaavi:


dhaavi:
Reviewed and approved by @vlabo.

PR #792, opened 2022-08-12
Add support for signed updates

dhaavi:


ppacher (dismissed):
LGTM, though, there are still a couple of `FIXME` comments

ppacher (commented):
LGTM

PR #802, opened 2022-08-19
Fix wayland options and disable restarting of UI components

dhaavi:
- Enable WaylandWindowDecorations feature for electron run on wayland
- Disable restarting on failure for UI components


ppacher (approved):

dhaavi:
Linter is broken.

PR #821, opened 2022-08-30
SPN Routing Improvements (and housekeeping)

dhaavi:
- Set tunneling options for (not) using community nodes
- Add new geoip metadata fields
- Improve logging of loaded intel data
- Move resolving unbreak filter list IDs to filterlists module
- Switch to zipfs lib for loading files from a zip
- Show update failed notification only after some failed tries
- Update Go, deps and linter


dhaavi:
Tests and Linter works locally.

ppacher (dismissed):

PR #855, opened 2022-09-15
Reset connection handling channel when firewall handler is stopped

dhaavi:


network/connection.go
```diff
@@ -590,9 +591,15 @@ func (conn *Connection) SetFirewallHandler(handler FirewallHandler) {
 }
 
 // StopFirewallHandler unsets the firewall handler and stops the handler worker.
+// The caller needs to hold a lock on the connection.
 func (conn *Connection) StopFirewallHandler() {
 	conn.firewallHandler = nil
+
+	// Signal the packet handler worker that it can stop.
 	conn.pktQueue <- nil
+
```
ppacher, line 600: just close the channel here and use `pkt, ok := <-pktQueue`

ppacher (changes requested):

network/connection.go
dhaavi: True. But I don't have to use the `ok` syntax, as it will return the zero value anyway, right?

ppacher: Right

ppacher (commented):
LGTM

PR #857, opened 2022-09-16
Feature/environment variables

vlabo:


process/process.go
```diff
@@ -209,6 +218,33 @@ func loadProcess(ctx context.Context, pid int) (*Process, error) {
...
+						log.Warningf(`invalid %s value: %q`, profileEnvironmentVariable, splitted[1])
+					}
+				} else {
+					log.Warningf(`empty value of %s`, profileEnvironmentVariable)
+				}
+				break
+			}
+		}
+	} else {
+		log.Errorf("failed to read environment variables: %q", err)
```
dhaavi, line 245: This and others: Please log to context and prefix the message with the module name. Eg.
`log.Tracer(ctx).Errorf("process: failed to read environment variables: %s", err)`

process/profile.go
```diff
@@ -82,10 +82,19 @@ func (p *Process) GetProfile(ctx context.Context) (changed bool, err error) {
...
 	}
 
-	// Get the (linked) local profile.
-	localProfile, err := profile.GetProfile(profile.SourceLocal, profileID, p.Path, false)
-	if err != nil {
-		return false, err
+	var localProfile *profile.Profile
+	if p.EnvironmentProfileID != "" && profileID == "" {
+		// Get the user specified profile.
+		localProfile, err = profile.GetProfile(profile.SourceEnvironmentDefined, p.EnvironmentProfileID, p.EnvironmentProfileID, false) // not setting process path since the profile an be used by multiple apps
```
dhaavi, line 88: Do use `EnvironmentProfileID` as the `linkedPath` - it's not a path. ;)
It is not required anyway.

process/process.go
```diff
@@ -209,6 +218,33 @@ func loadProcess(ctx context.Context, pid int) (*Process, error) {
...
 	process.ParentPid = int(ppid)
 
+	// Get all environment variables
+	envVariables, err := pInfo.EnvironWithContext(ctx)
+
+	// Not important enough to stop execution of the function
+	if err == nil {
+		// Try to find PORTMASTER_PROFILE variable
+		for _, envVar := range envVariables {
+			if strings.HasPrefix(envVar, profileEnvironmentVariable) {
```
dhaavi, line 228: Should we be matching for `PORTMASTER_PROFILE=`?
Right now, this would match anything starting with `PORTMASTER_PROFILE`.

process/process.go
```diff
@@ -209,6 +218,33 @@ func loadProcess(ctx context.Context, pid int) (*Process, error) {
...
+				// Found. Check validity and set to process
+				splitted := strings.SplitN(envVar, "=", 2)
```
dhaavi, line 230: It might be easier to just use `strings.TrimPrefix()` if we already know what it is.

process/process.go
```diff
@@ -55,6 +61,9 @@ type Process struct {
 	// profile holds the layered profile based on the primary profile.
 	profile *profile.LayeredProfile
 
+	// Holds the ID of a profile set from the PORTMASTER_PROFILE enlivenment variable or empty if not available
```
dhaavi, line 64: Typo at `enlivenment`.

profile/get.go
```diff
@@ -142,7 +142,7 @@ func getProfile(scopedID string) (profile *Profile, err error) {
 
 // findProfile searches for a profile with the given linked path. If it cannot
 // find one, it will create a new profile for the given linked path.
-func findProfile(linkedPath string) (profile *Profile, err error) {
+func findProfile(id string, linkedPath string) (profile *Profile, err error) {
```
dhaavi, line 145: `findProfile` should only be used for path based profiles.

profile/get.go
```diff
@@ -72,7 +72,7 @@ func GetProfile(source profileSource, id, linkedPath string, reset bool) ( //nol
 
 		// Get from database.
 		if !reset {
-			profile, err = findProfile(linkedPath)
+			profile, err = findProfile(id, linkedPath)
 			// Check if the profile is special and needs a reset.
 			if err == nil && specialProfileNeedsReset(profile) {
 				profile = getSpecialProfile(id, linkedPath)
```
dhaavi, line 78: L90-100:

There seems to be a bug already in here, as line 89 should call `New()` with the given source.

This is also the point where we would need to add checks on which profiles can be created automatically in which case if we follow along with my suggestion below. An additional flag to the function of `idByUser` could specify whether the profile ID was supplied by the user and may only be automatically created in the `user` scope.
Also, we can probably ditch the checks on line 111 for now, until we start using layered profiles.

dhaavi (changes requested):

profile/profile.go
```diff
@@ -26,11 +26,12 @@ type profileSource string
...
-	SourceSpecial    profileSource = "special" // specials (read-only)
-	SourceNetwork    profileSource = "network"
-	SourceCommunity  profileSource = "community"
-	SourceEnterprise profileSource = "enterprise"
+	SourceLocal              profileSource = "local"   // local, editable
+	SourceSpecial            profileSource = "special" // specials (read-only)
+	SourceNetwork            profileSource = "network"
+	SourceCommunity          profileSource = "community"
+	SourceEnterprise         profileSource = "enterprise"
+	SourceEnvironmentDefined profileSource = "environment-defined" // defined from environment variable
```
dhaavi, line 34: Ok, so GitHub ditched my longest comment here. Ugh. Here we go again...

I thought about the whole use case of this feature and how it interacts with other features.
If we strictly scope profiles from this feature to a separate source, it will not be able to interact with other features.

For example, @ppacher suggested a feature, where one could provision profiles/settings by placing them in a configured directory on disk. Portmaster would then search the profiles and apply them if one matched. These could live in a scope name `provision-dir`.
These profiles however, would be inaccessible to use via the environment variable.

My suggestion is:
- Create a new generic source called `user`, which holds profiles created by the user. (Also, from the UI in the future.)
- If a profile ID is specified via the environment variable and no `/` is in the ID, we assume the source `user`.
- If there is a source defined, we load the profile from there, BUT we would not automatically create one there, but fail instead.

I think this is a nice small change that will help future-proof this feature.
What are your thoughs @vlabo and @ppacher?

ppacher: I really like the idea but IMHO we could leave the "environment-defined" source if we just add the support to specify the profile source in the env-variable. I.e. `my-custom-profile` would create a profile with the ID `env-var/my-custom-profile` while `file/my-custom-profile` would try to load the profile from the provisioning directory and fail if it does not exist. Likewise, using `local/0000-0000-0000-0000` would try to load the profile generated by the PM or fail otherwise. This way we don't need to introduce a somewhat one-for-all `user` source

ppacher: Also we should probably add a developer/advanced setting that enables the use of environment variables for profile selection?

dhaavi: Let's go with:
- `env-var` source for profiles created from the environment variable (if not source is defined)
- Create an advanced setting that enabled environment variable profile selection
- profiles from the env var may only be created in the `env-var` scope, if loading from another source fails, it fails and all connection should be blocked (as the profile is missing).

PR #918, opened 2022-10-03
Add support for variable profile matching

dhaavi:
- Move from LinkedPath to Fingerprints and PresentationPath
- Collect env, tags and and matching path for processes
- Add first tag handlers for svchost, app image, net
- Use new profile functions in firewall


process/tags/svchost_windows.go
```diff
@@ -82,6 +82,8 @@ func (h *SVCHostTagHandler) CreateProfile(p *process.Process) *profile.Profile {
 			return profile.New(&profile.Profile{
 				Source:              profile.SourceLocal,
 				Name:                "Windows Service: " + osdetail.GenerateBinaryNameFromPath(tag.Value),
+				Icon:                `C:\Windows\System32\@WLOGO_48x48.png`,
+				IconType:            profile.IconTypeFile,
 				UsePresentationPath: false,
```
ppacher: That one is not yet supported by the UI since that would require electron to actually load that file. `blob` and `database` are supported. Maybe we should add `file` support using a Portmaster API?

ppacher (approved):

process/tags/interpreter_unix.go
```diff
@@ -0,0 +1,220 @@
...
+		// so it may be better to scan the whole command line for a path to a UTF8
+		// file and use that one.
+		if !fileMustBeUTF8(filePath) {
+			return
+		}
+
+		p.Tags = append(p.Tags, profile.Tag{
+			Key:   matched.ID,
+			Value: filePath,
+		})
```
dhaavi, line 209: Should we set this path as the MatchingPath?

process/tags/interpreter_unix.go
```diff
@@ -0,0 +1,220 @@
...
+			return
+		}
+
+		filePath := args[1]
+		if !filepath.IsAbs(filePath) {
+			filePath = filepath.Join(
+				p.Cwd,
+				filePath,
+			)
+		}
```
dhaavi, line 197: Using `filePath` and `filepath` makes this hard to read.

process/tags/interpreter_unix.go
```diff
@@ -0,0 +1,220 @@
...
+
+		return
+	}
+
+	// we know that this process is interpreted by some known interpreter but args[0]
+	// does not contain the path to the interpreter.
+	p.Tags = append(p.Tags, profile.Tag{
+		Key:   matched.ID,
+		Value: args[0],
+	})
```
dhaavi, line 219: Should we set this path as the MatchingPath?

process/tags/interpreter_unix.go
```diff
@@ -0,0 +1,220 @@
...
+
+			// if arg0 is the interpreter name itself strip it away
+			// and use the next one
+			if it.Regex.MatchString(args[0]) && len(args) > 1 {
+				args = args[1:]
+			}
+
+			return profile.New(&profile.Profile{
+				Source:              profile.SourceLocal,
+				Name:                fmt.Sprintf("%s: %s", it.Name, args[0]),
```
dhaavi, line 145: Is it simpler to use (a variation of) the tag value?

package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "workspace",
+  "lockfileVersion": 2,
+  "requires": true,
+  "packages": {}
+}
```
dhaavi, line 6: Don't think this belongs here.

PR #923, opened 2022-10-10
Add experimental support to inherit profiles from parent processes

ppacher:
Building on top-of the new profile matching feature this PR adds an experimental setting to enable profile inheritance for processes.

If enabled, the algorithm to find a matching profile is as follows:

1. Find profile and highest fingerprint score for the current process
2. Get fingerprint matching data for the parent process
3. Find profile and highest fingerprint score for that parent
4. If the fingerprint score is higher, use the parent profile instead.
5. Repeat steps 2-4 until there is no more parent process.

This change effectively allows applying a profile to all sub-processes of a process. For example, this may be useful in the following scenarios:

- A profile for the vscode (matched by cmdline or binary) currently is not correctly applied to the electron sub-processes of that profile. With the above changes, a cmdline match for vscode will have a higher priority and thus force all electron sub-processes to have the same profile applied.

- Shell or interpreter scripts will get a profile with a script-tag assigned. Thus, all sub-processes executed by this script will have the correct profile applied, i.e. a shell script that calls `curl` now receives the correct profile (that on the script file)


ppacher:
Needs rebasing

ppacher:
Will be re-created

PR #925, opened 2022-10-10
Make sure worst verdict is updated after final verdict is set

vlabo:


dhaavi:
I don't think this changes anything - it also works for me. Let's discuss this.

vlabo:
Not a fix. Was not able to reproduce the issue

PR #929, opened 2022-10-11
Fix fs error handling and other things

dhaavi:
- Fix fs error handling
- Use per-user lock file when unlocking
- Use interception config only when ready
- Improve logging and context usage
- Check if active profile is outdated before using
- Update deps


vlabo (approved):

profile/get.go
```diff
@@ -221,7 +221,7 @@ profileFeed:
 	}
 
 	// Check if this profile is already active and return the active version instead.
-	if activeProfile := getActiveProfile(profile.ScopedID()); activeProfile != nil {
+	if activeProfile := getActiveProfile(profile.ScopedID()); activeProfile != nil && !activeProfile.IsOutdated() {
 		return activeProfile, nil
 	}
```
ppacher, line 226: wouldn't it be better to still return the outdated profile since that one will get refreshed anyway? If we return a new profile here we will have two instances of that profile in RAM which will likely result in some strange behavior?

ppacher (commented):
LGTM, One question

profile/get.go
dhaavi: This is only used when loading a profile from disk and an active one was not found.
If this new profile is made active, the old active one is marked outdated.
This fixed an issue where it would always continue to use the old profile instead of loading changes made in the UI.

PR #967, opened 2022-11-02
Remove the glue library for the kernel extension on windows

vlabo:


firewall/interception/windowskext/service.go
```diff
@@ -0,0 +1,87 @@
...
+	"fmt"
+	"syscall"
+
+	"golang.org/x/sys/windows"
+)
+
+func createService(manager windows.Handle, portmasterKextPath *uint16) (windows.Handle, error) {
+	u16filename, err := syscall.UTF16FromString(driverName)
+	if err != nil {
+		return 0, fmt.Errorf("Bad service: %s", err)
```
dhaavi, line 16: not sure if `Bad service` is a good error description.
How about `failed to convert name to UTF16 string: %w`?

firewall/interception/windowskext/service.go
```diff
@@ -0,0 +1,87 @@
...
+	service, err = windows.CreateService(manager, &u16filename[0], &u16filename[0], windows.SERVICE_ALL_ACCESS, windows.SERVICE_KERNEL_DRIVER, windows.SERVICE_DEMAND_START, windows.SERVICE_ERROR_NORMAL, portmasterKextPath, nil, nil, nil, nil, nil)
+	if err != nil {
+		return 0, err
+	}
+
+	return service, nil
+}
+
+func driverInstall(portmasterKextPath string) (windows.Handle, error) {
+	u16kextPath, _ := syscall.UTF16FromString(portmasterKextPath)
```
dhaavi, line 34: Could this fail?

firewall/interception/windowskext/service.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// Try to create the service. Retry if it fails.
+	var service windows.Handle
+retryLoop:
+	for i := 0; i < 3; i++ {
+		service, err = createService(manager, &u16kextPath[0])
+		if err == nil {
+			break retryLoop
+		}
+	}
```
dhaavi, line 50: Please document _why_ you are retrying

firewall/interception/windowskext/service.go
```diff
@@ -0,0 +1,87 @@
...
+
+	if err != nil {
+		return 0, fmt.Errorf("Failed to create service: %s", err)
```
dhaavi, line 53: errors should always start lower case.
Also, please wrap the error with `%w`.

firewall/interception/windowskext/service.go
```diff
@@ -0,0 +1,87 @@
...
+			_ = windows.CloseServiceHandle(service)
+			service = 0
+		}
+	}
+
+	return service, nil
+}
+
+func openDriver(filename string) (windows.Handle, error) {
+	u16filename, _ := syscall.UTF16FromString(filename)
```
dhaavi, line 75: Could this fail?

firewall/interception/windowskext/kext.go
```diff
@@ -25,108 +25,57 @@ var (
...
+	_ = windows.CloseServiceHandle(service)
 
-	rc, _, lastErr := kext.start.Call(
-		uintptr(unsafe.Pointer(&charArray[0])),
-	)
-	if rc != windows.NO_ERROR {
-		return formatErr(lastErr, rc)
+	// driver was not installed
+	if err != nil {
+		return fmt.Errorf("Failed to start the kext service: %s %q", err, filename)
```
dhaavi, line 78: Error messages should always start lower case.
Also, please use `%w` to wrap the error.
I would include the filename in the message and put the error last, as it may be long. This will make it easier to make sense of the message.

firewall/interception/windowskext/kext.go
```diff
@@ -25,108 +25,57 @@ var (
...
-	new.init, err = new.dll.FindProc("PortmasterInit")
-	if err != nil {
-		return fmt.Errorf("could not find proc PortmasterStart in dll: %s", err)
-	}
-	new.start, err = new.dll.FindProc("PortmasterStart")
+	// initialize and start driver service
+	service, err := driverInstall(driverPath)
 	if err != nil {
-		return fmt.Errorf("could not find proc PortmasterStart in dll: %s", err)
+		return fmt.Errorf("Failed to start service: %s", err)
```
dhaavi, line 66: Error messages should always start lower case.
Also, please use `%w` to wrap the error.

firewall/interception/windowskext/service.go
```diff
@@ -0,0 +1,87 @@
...
+		err = windows.GetLastError()
+		if err != windows.ERROR_SERVICE_ALREADY_RUNNING {
+			// Failed to start service; clean-up:
+			var status windows.SERVICE_STATUS
+			_ = windows.ControlService(service, windows.SERVICE_CONTROL_STOP, &status)
+			_ = windows.DeleteService(service)
+			_ = windows.CloseServiceHandle(service)
+			service = 0
+		}
+	}
```
dhaavi, line 69: This error (from L57) is not returned.

firewall/interception/windowskext/service.go
```diff
@@ -0,0 +1,87 @@
...
+	err = windows.StartService(service, 0, nil)
+
+	if err != nil {
+		err = windows.GetLastError()
+		if err != windows.ERROR_SERVICE_ALREADY_RUNNING {
+			// Failed to start service; clean-up:
+			var status windows.SERVICE_STATUS
+			_ = windows.ControlService(service, windows.SERVICE_CONTROL_STOP, &status)
+			_ = windows.DeleteService(service)
+			_ = windows.CloseServiceHandle(service)
```
dhaavi, line 66: Why is the service stopped and removed? Is this only our reference to it?
In any case this seems unusual and should be documented.
We want to know why we are doing this next time we look at it.

firewall/interception/windowskext/kext.go
```diff
@@ -25,108 +25,57 @@ var (
...
+	// open the driver
+	kextHandle, err = openDriver(filename)
 
-	// convert to C string
-	charArray := make([]byte, len(kext.driverPath)+1)
-	copy(charArray, []byte(kext.driverPath))
-	charArray[len(charArray)-1] = 0 // force NULL byte at the end
+	// close the service handles
+	_ = windows.DeleteService(service)
+	_ = windows.CloseServiceHandle(service)
```
dhaavi, line 74: Why is the service removed? Is this only our reference to it?
In any case this seems unusual and should be documented.
We want to know why we are doing this next time we look at it.

firewall/interception/windowskext/kext.go
```diff
@@ -142,9 +91,14 @@ func Stop() error {
 	}
 	ready.UnSet()
 
-	rc, _, lastErr := kext.stop.Call()
-	if rc != windows.NO_ERROR {
-		return formatErr(lastErr, rc)
+	err := closeDriver(kextHandle)
```
dhaavi, line 94: With how we handle this now, maybe it does not make sense to check for `!ready.IsSet()` and just check if `kextHandle` is valid and should be closed instead.

firewall/interception/windowskext/kext.go
```diff
@@ -142,9 +91,14 @@ func Stop() error {
...
+	if err != nil {
+		log.Errorf("winkext: failed to close the handle: %s", err)
+	}
+
+	_, err = exec.Command("sc", "stop", driverName).Output() // This is a question of taste, but it is a robust and solid solution
+	if err != nil {
+		log.Errorf("winkext: failed to stop the service: %q", err)
 	}
```
dhaavi, line 102: Why not use the service manager here?
What makes it robust and solid?
I find it weird to use one way to start and another to stop it.

firewall/interception/windowskext/kext.go
```diff
@@ -170,19 +121,22 @@ func RecvVerdictRequest() (*VerdictRequest, error) {
 		time.Sleep(100 * time.Microsecond)
 	}
 
-	// timestamp := time.Now()
-	rc, _, lastErr := kext.recvVerdictRequest.Call(
-		uintptr(unsafe.Pointer(new)),
-	)
-	// log.Tracef("winkext: getting verdict request took %s", time.Now().Sub(timestamp))
+	timestamp := time.Now()
```
dhaavi, line 124: I would recommend using the following pattern for this kind of debugging:
```
started := time.Now()
defer log.Tracef("winkext: getting verdict request took %s", time.Since(started))
```
This keeps everything together.

firewall/interception/windowskext/kext.go
```diff
@@ -156,9 +110,6 @@ func RecvVerdictRequest() (*VerdictRequest, error) {
 	if !ready.IsSet() {
 		return nil, ErrKextNotReady
 	}
-
-	new := &VerdictRequest{}
-
 	// wait for urgent requests to complete
```
dhaavi, line 113: This was used to prioritize setting verdicts over receiving verdicts. Do we still need this? Ie. is access to the driver still locked and synchronous?

firewall/interception/windowskext/kext.go
```diff
@@ -199,17 +153,16 @@ func SetVerdict(pkt *Packet, verdict network.Verdict) error {
...
 	atomic.AddInt32(urgentRequests, 1)
-	// timestamp := time.Now()
-	rc, _, lastErr := kext.setVerdict.Call(
-		uintptr(pkt.verdictRequest.id),
-		uintptr(verdict),
-	)
-	// log.Tracef("winkext: settings verdict for packetID %d took %s", packetID, time.Now().Sub(timestamp))
+	data := asByteArray(&verdictInfo)
+	_, err := deviceIoControlWrite(kextHandle, IOCTL_SET_VERDICT, data)
 	atomic.AddInt32(urgentRequests, -1)
```
dhaavi, line 162: Unrelated: This should be called with a defer right after the AddInt32.

dhaavi (changes requested):
This is a great improvement! Looking forward to what we will be able to do with this in the future!

PR #995, opened 2022-11-11
Build for all platforms in CI test

dhaavi:


dhaavi:
Fixes + texts reviewed by @davegson

PR #1014, opened 2022-11-24
Fix DoH base64 encoding

vlabo:


resolver/resolver-https.go
```diff
@@ -94,6 +94,10 @@ func (hr *HTTPSResolver) Query(ctx context.Context, q *Query) (*RRCache, error)
 		_ = resp.Body.Close()
 	}()
 
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("http response code %s", resp.Status)
```
dhaavi, line 98: Please be more descriptive and use a verb-form for the error. Eg: `http request failed with %s`

dhaavi:
```suggestion
		return nil, fmt.Errorf("http request failed with %s", resp.Status)
```

dhaavi (changes requested):

dhaavi (approved):

PR #1134, opened 2023-03-16
Fix traceroute locating, improvements

dhaavi:
- Fix traceroute location checking
- Improve network proximity thresholds
- Warn when query fails


vlabo (approved):

dhaavi:
Linter is broken in CI, works locally.

PR #1290, opened 2023-08-18
Fix connection state handling and improve DNS request attribution

dhaavi:
- Remove check for self-communication (allowed by default)
- Use DNS request connections to attribute DNS requests
- Update network state tables if state is older than given packet
- Fix network history clear API
- Update special profile upgrade date
- Improve bandwidth update logging
- Recompile ebpf programs


network/dns.go
```diff
@@ -38,6 +42,82 @@ const (
...
+	if !ok {
+		return nil, false
+	}
+
+	// Get and return
+	dnsRequestConnectionsLock.RLock()
+	defer dnsRequestConnectionsLock.RUnlock()
+
+	conn, ok = dnsRequestConnections[key]
+	return
```
ppacher, line 89: Avoid naked returns

network/dns.go
```diff
@@ -38,6 +42,82 @@ const (
...
+}
+
+// deleteDNSRequestConnection removes a connection from the dns request connections.
+func deleteDNSRequestConnection(packetInfo *packet.Info) { //nolint:unused,deadcode
```
ppacher, line 93: Why nolint:deadcode?

process/process.go
```diff
@@ -112,6 +112,16 @@ func (p *Process) IsIdentified() bool {
 	}
 }
 
+// IsLocal returns whether the process has been identified as a local process.
```
ppacher, line 115: The comment is a bit misleading, bow can a process not be "local"

ppacher (changes requested):
LGTM, just a few minor things

PR #1311, opened 2023-09-06
Various QoL improvments

ppacher:


dhaavi (approved):
Tests are failing.

core/config.go
```diff
@@ -13,6 +13,8 @@ var (
 
 	CfgNetworkServiceKey      = "core/networkService"
 	defaultNetworkServiceMode bool
+
+	CfgLocaleKey = "core/localeID"
```
dhaavi, line 17: Would suggest only `core/locale`.

core/config.go
```diff
@@ -41,5 +43,32 @@ func registerConfig() error {
...
+		PossibleValues: []config.PossibleValue{
+			{
+				Name:  "en-GB",
+				Value: "en-GB",
+			},
+			{
+				Name:  "en-US",
+				Value: "en-US",
+			},
+		},
```
dhaavi, line 63: Would suggest:

Name: `Date and Number Format`
Options: `American Format`, `European Format`

This is a lot easier to understand and not _too_ wrong for the start. ;)

Later this could evolve into language+format, as the locale also includes the language.

dhaavi (changes requested):

PR #1326, opened 2023-09-22
Add first set of import/export APIs

dhaavi:


sync/settings.go
```diff
@@ -0,0 +1,305 @@
...
+		Parameters: []api.Parameter{{
+			Method:      http.MethodGet,
+			Field:       "from",
+			Description: "Specify where to export from.",
+		}},
+		BelongsTo: module,
+		DataFunc:  handleExportSettings,
+	}); err != nil {
+		return err
+	}
```
ppacher, line 52: We don't set the correct content type for this API endpoint

sync/settings.go
```diff
@@ -0,0 +1,305 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Make some yummy yaml.
+	yamlData, err := yaml.Marshal(export)
+	if err != nil {
+		return nil, fmt.Errorf("%w: failed to marshal to yaml: %s", ErrExportFailed, err)
+	}
+
```
ppacher, line 114: It would be awesome to check the "Accept" header here and return the expected content type. i.e `application/json` , `text/yaml` or `text/x-yaml`

sync/settings.go
```diff
@@ -0,0 +1,305 @@
...
+	// Check if there any changed settings.
+	if len(settings) == 0 {
+		return nil, ErrUnchanged
+	}
+
+	// Expand config to hierarchical form.
+	settings = config.Expand(settings)
+
```
ppacher, line 191: When exporting a single setting we do that "flattened" but for the global setting we do that expanded. This seems a bit inconsistent to me

sync/settings.go
```diff
@@ -0,0 +1,305 @@
...
+			}
+		}
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+	if checked < len(settings) {
+		return nil, fmt.Errorf("%w: the export contains unknown settings", ErrInvalidImport)
+	}
```
ppacher, line 243: This will break the plugin system when it eventually get's merged

sync/settings.go
```diff
@@ -0,0 +1,305 @@
...
+	// Import settings into profile.
+	if r.Reset {
+		p.Config = config.Expand(settings)
+	} else {
+		flattenedProfileConfig := config.Flatten(p.Config)
+		for k, v := range settings {
+			flattenedProfileConfig[k] = v
+		}
+		p.Config = config.Expand(flattenedProfileConfig)
+	}
```
ppacher, line 296: Same as above, the whole flattening and expanding of config options seems like useless overhead

sync/settings.go
dhaavi: I would expect the plugins to be already loaded when settings are imported.

We could however, treat this as a warning instead - but we can also "weaken" this later on.

sync/settings.go
dhaavi: The problem here is with the checksum/signature that we want to add to the export.

I only see a clean way to do this with yaml comments, but not with json...

dhaavi: Support for yaml and json is now done.

PR #1363, opened 2023-11-15
Add export and import for profiles

dhaavi:


sync/profile.go
```diff
@@ -1,45 +1,440 @@
...
+	Internal   bool       `json:"internal,omitempty"`
+}
+
+// ProfileIcon represents a profile icon.
+type ProfileIcon struct {
+	Type  profile.IconType `json:"type"`
+	Value string           `json:"value"`
+}
+
+// ProfileIcon represents a profile fingerprint.
```
ppacher, line 51: Comment is for `ProfileIcon` not for `ProfileFingerprint`

sync/profile.go
```diff
@@ -1,45 +1,440 @@
...
+				ValidateOnly: q.Has("validate"),
+				RawExport:    string(ar.InputData),
+				RawMime:      ar.Header.Get("Content-Type"),
+			},
+			AllowUnknown: q.Has("allowUnknown"),
+			AllowReplace: q.Has("allowReplace"),
+		}
+	} else {
+		request = &ProfileImportRequest{}
+		if err := json.Unmarshal(ar.InputData, request); err != nil {
```
ppacher, line 184: Allocating an empty struct for `request` is not required, json.Unmarshal will do it anyway

sync/profile.go
```diff
@@ -1,45 +1,440 @@
...
+func ExportProfile(scopedID string) (*ProfileExport, error) {
+	// Get Profile.
+	r, err := db.Get(profile.ProfilesDBPath + scopedID)
+	if err != nil {
+		return nil, fmt.Errorf("%w: failed to find profile: %w", ErrTargetNotFound, err)
+	}
+	p, err := profile.EnsureProfile(r)
+	if err != nil {
+		return nil, fmt.Errorf("%w: failed to load profile: %w", ErrExportFailed, err)
+	}
```
ppacher, line 220: Shouldn't the profile get locked here?

sync/settings.go
```diff
@@ -334,3 +298,48 @@ func ImportSettings(r *SettingsImportRequest) (*ImportResult, error) {
...
+
+	// Validate config and gather some metadata.
+	var checked int
+	err = config.ForEachOption(func(option *config.Option) error {
+		// Check if any setting is set.
+		// TODO: Fix this - it only checks for global settings.
+		// if r.Reset && option.IsSetByUser() {
+		// 	result.ReplacesExisting = true
+		// }
+
```
ppacher, line 313: Is this something that needs to be sorted out before we release the import/export feature?

ppacher (changes requested):

sync/profile.go
dhaavi: > Unmarshal parses the JSON-encoded data and stores the result in the value pointed to by v. If v is nil or not a pointer, Unmarshal returns an InvalidUnmarshalError.

sync/profile.go
dhaavi: This is parsing a raw entry from the DB. Nobody has can have a lock on this.

PR #1370, opened 2023-11-24
Remove network rating / security level system, add migrations

dhaavi:


dhaavi:
Requires https://github.com/safing/portbase/pull/194

PR #1426, opened 2024-01-22
Monitor System DNS client queries

dhaavi:
This feature adds monitoring capabilities for the system DNS client.

How?
Windows: Windows Events subcription﻿ on the DNS Client
Linux: DBUS monitoring of systemd-resolved

This will allow Portmaster to disable the Secure DNS, while keeping DNS insights.

### DIFF
--- a/systemdns/dnsevtlog/events.go
+++ b/systemdns/dnsevtlog/events.go
@@ -1,0 +1,139 @@
+//go:build windows
+// +build windows
+
+package dnsevtlog
+
+// This code is copied from Promtail v1.6.2-0.20231004111112-07cbef92268a with minor changes.
+
+import (
+	"fmt"
+	"sync"
+	"time"
+
+	"golang.org/x/sys/windows"
+
+	"github.com/safing/portbase/log"
+	"github.com/safing/portmaster/systemdns/dnsevtlog/win_eventlog"
+)
+
+type Subscription struct {
+	subscription win_eventlog.EvtHandle
+	fetcher      *win_eventlog.EventFetcher
+
+	eventLogName  string
+	eventLogQuery string
+
+	ready bool
+	done  chan struct{}
+	wg    sync.WaitGroup
+	err   error
+}
+
+// NewSubscription create a new windows event subscriptions.
+func NewSubscription() (*Subscription, error) {
+	sigEvent, err := windows.CreateEvent(nil, 0, 0, nil)
+	if err != nil {
+		return nil, err
+	}
+	defer windows.CloseHandle(sigEvent)
+
+	t := &Subscription{
+		eventLogName:  "Microsoft-Windows-DNS-Client/Operational",
+		eventLogQuery: "*",
+		done:          make(chan struct{}),
+		fetcher:       win_eventlog.NewEventFetcher(),
+	}
+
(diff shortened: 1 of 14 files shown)


systemdns/dnsevtlog/events.go
```diff
@@ -0,0 +1,139 @@
...
+		return nil, fmt.Errorf("error subscribing to windows events: %w", err)
+	}
+	t.subscription = subsHandle
+
+	return t, nil
+}
+
+// loop fetches new events and send them to via the Loki client.
+func (t *Subscription) ReadWorker() {
+	t.ready = true
```
ppacher, line 58: Comment needs to be updated

systemdns/dnsevtlog/events.go
```diff
@@ -0,0 +1,139 @@
...
+	err   error
+}
+
+// NewSubscription create a new windows event subscriptions.
+func NewSubscription() (*Subscription, error) {
+	sigEvent, err := windows.CreateEvent(nil, 0, 0, nil)
+	if err != nil {
+		return nil, err
+	}
+	defer windows.CloseHandle(sigEvent)
```
ppacher, line 38: Why do we need this? `sigEvent` is not used anymore in the function body.

systemdns/dnsevtlog/events.go
```diff
@@ -0,0 +1,139 @@
...
+	}()
+
+	for {
+
+	loop:
+		for {
+			// fetch events until there's no more.
+			events, handles, err := t.fetcher.FetchEvents(t.subscription, 1033) // 1033: English
+			if err != nil {
+				if err != win_eventlog.ERROR_NO_MORE_ITEMS {
```
ppacher, line 74: should be `errors.Is`

systemdns/dnsevtlog/events.go
```diff
@@ -0,0 +1,139 @@
...
+)
+
+type Subscription struct {
+	subscription win_eventlog.EvtHandle
+	fetcher      *win_eventlog.EventFetcher
+
+	eventLogName  string
+	eventLogQuery string
+
+	ready bool
```
ppacher, line 26: ready is never read anywhere so I guess we can drop it.

systemdns/dnsevtlog/events.go
```diff
@@ -0,0 +1,139 @@
...
+// loop fetches new events and send them to via the Loki client.
+func (t *Subscription) ReadWorker() {
+	t.ready = true
+	t.wg.Add(1)
+	interval := time.NewTicker(time.Second)
+	defer func() {
+		t.ready = false
+		t.wg.Done()
+		interval.Stop()
+	}()
```
ppacher, line 65: I guess `ReadWorker` is expected to run in a goroutine. so access to `read` is unprotected if it's launched multiple times.
Also, `wg.Add(1)` should be called before the go-routine is launched since there could be a race condition when starting the `ReadWorker` and immediatley calling `Stop` which could cause a panic. See https://pkg.go.dev/sync#WaitGroup.Add.

Better add a dedicated `Start()` routine that adds to the wg and kicks of the goroutine

ppacher (changes requested):

PR #1439, opened 2024-02-09
Improve self-check, port names, exe icons, online checks

dhaavi:
- Skip self-check if device is offline
- Add ping.archlinux.org to connectivity check domains
- Stop using dashes for port names, as they collide with port range dashes
- Improve finding windows exe icon

### DIFF
--- a/network/reference/ports.go
+++ b/network/reference/ports.go
@@ -14,8 +14,8 @@ var (
 		25:  "SMTP",
 		43:  "WHOIS",
 		53:  "DNS",
-		67:  "DHCP-SERVER",
-		68:  "DHCP-CLIENT",
+		67:  "DHCP_SERVER",
+		68:  "DHCP_CLIENT",
 		69:  "TFTP",
 		80:  "HTTP",
 		110: "POP3",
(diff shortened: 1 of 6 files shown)


network/reference/ports.go
```diff
@@ -14,8 +14,8 @@ var (
 		25:  "SMTP",
 		43:  "WHOIS",
 		53:  "DNS",
-		67:  "DHCP-SERVER",
-		68:  "DHCP-CLIENT",
+		67:  "DHCP_SERVER",
+		68:  "DHCP_CLIENT",
```
ppacher, line 18: Isn't that a breaking change? Rules that use the old port-names will be broken now. Should we do a migration for that?

profile/binmeta/find_windows.go
```diff
@@ -111,5 +88,31 @@ func getIconAndNamefromRSS(ctx context.Context, binPath string) (png []byte, nam
...
+	rss.WalkType(winres.RT_GROUP_ICON, func(resID winres.Identifier, langID uint16, _ []byte) bool {
+		icon, iconErr = rss.GetIconTranslation(resID, langID)
+		return iconErr != nil
+	})
+	if iconErr != nil {
+		return nil, name, fmt.Errorf("failed to get icon: %w", err)
+	}
+	if icon == nil {
+		return nil, name, errors.New("no icon in resources")
+	}
```
ppacher, line 105: we could return `fs.ErrNotFound` here

ppacher (dismissed):

network/reference/ports.go
dhaavi: For name to number conversion both versions are supported.

PR #1469, opened 2024-03-27
Add support for --allowed-clients parameter to whitelist binaries that are allowed to talk to the Portmaster API

ppacher:


service/firewall/api.go
```diff
@@ -164,6 +165,15 @@ func authenticateAPIRequest(ctx context.Context, pktInfo *packet.Info) (retry bo
...
 				// Check if the requesting process is in database root / updates dir.
 				if realPath, err := filepath.EvalSymlinks(proc.Path); err == nil {
+
+					// check if the client has been allowed by flag
+					if slices.Contains(allowedClients, realPath) {
+						log.Infof("filter: access to portmaster api allowed for configured client: %s", realPath)
+						return false, nil
+					} else if len(allowedClients) > 0 {
+						log.Warningf("filter: process is not in the allowed clients list: %s (list=%s)", realPath, allowedClients)
+					}
```
dhaavi, line 175: This will work, but I think everything should already be logged elsewhere.
Especially if we were to match a parent process, it would still print a warning that the child did not match.

dhaavi (changes requested):

PR #1498, opened 2024-04-08
Refactor: Edit build command to match angular.json

Exitium-DEV:
The current advised build command does not export the correct environment, resulting in a developer environment being exported, even at production-level.

This causes an issue when we start accessing the web server from a hostname that is not localhost, where the websocket cannot establish a connection to 127.0.0.1 instead of window.location.host


dhaavi:
As mentioned on Discord, I think this might get resolved with the new build system, which I am currently working on creating the first release with it.

PR #1506, opened 2024-04-12
[fix] (Windows;performance) Ineffective break statement.

stenya:
Break out of the outer loop.


dhaavi (approved):

dhaavi:
Hey @stenya, good catch, thanks again!

PR #1521, opened 2024-04-26
fix: close file

testwill:


dhaavi:
Hey @testwill, thanks for the fix - great find!

dhaavi (approved):

PR #1641, opened 2024-08-01
Fix Tauri cmdline args

dhaavi:
- [cmd/start] Compile cmdline args after possibly upgrading to new UI
- Update ignore files


dhaavi (commented):
LGTM

PR #1820, opened 2025-02-25
Add SQLite database storage backend (v1.x)

dhaavi:
- **Add SQLite database storage backend**
- **Switch core and cache databases to use sqlite when bbold db is not present**


dhaavi:
Remaining linter errors are very low impact.
Tests are working now.

PR #1846, opened 2025-03-26
updater fix

stenya:
[updates] fix: use correct binaries index name
[updates] fix: Initial update failed due to the published date of the actual index being in the past


stenya:
Canceled

PR #1849, opened 2025-03-28
Fix/updater fix 2

stenya:
[updates] fix: use correct binaries index name
[updates] fix: update from locally generated index (timestamps and downgrade)

### DIFF
--- a/service/configure/updates.go
+++ b/service/configure/updates.go
@@ -5,6 +5,9 @@ import (
 )
 
 var (
+	DefaultBinaryIndexName = "Portmaster Binaries"
+	DefaultIntelIndexName  = "intel"
+
 	DefaultStableBinaryIndexURLs = []string{
 		"https://updates.safing.io/stable.v3.json",
 	}
--- a/service/updates/upgrade.go
+++ b/service/updates/upgrade.go
@@ -73,6 +73,10 @@ func (u *Updater) upgradeMoveFiles(downloader *Downloader) error {
 			if slices.Contains(u.cfg.Ignore, file.Name()) {
 				continue
 			}
+			// ignore PurgeDirectory itself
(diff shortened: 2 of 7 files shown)


service/configure/updates.go
```diff
@@ -5,6 +5,9 @@ import (
 )
 
 var (
+	DefaultBinaryIndexName = "Portmaster Binaries"
+	DefaultIntelIndexName  = "intel"
```
dhaavi, line 9: The current name from https://updates.safing.io/intel.v3.json is "Portmaster Intel", in case you wanted to sync with that.

service/updates/upgrade.go
```diff
@@ -73,6 +73,10 @@ func (u *Updater) upgradeMoveFiles(downloader *Downloader) error {
 			if slices.Contains(u.cfg.Ignore, file.Name()) {
 				continue
 			}
+			// ignore PurgeDirectory itself
+			if strings.EqualFold(u.cfg.PurgeDirectory, filepath.Join(u.cfg.Directory, file.Name())) {
+				continue
+			}
```
dhaavi, line 79: Not sure why you need this.
The directories should not / must not overlap.

dhaavi (commented):
LGTM, 2 comments.

service/updates/upgrade.go
stenya: This is required because there was an issue with copying — it tried to copy the PurgeDirectory into itself, as I recall.

This change fixes this error:

`2025-03-26 17:54:21.859 ERR mgr\workermgr.go:219 worker failed manager="Binary Updater" worker=upgrader err="udpating failed: upgrade failed, but recovery was successful: failed to move current file upgrade_obsolete_binaries to purge dir: move failed, copy+delete fallback failed: read src file: read C:\\Program Files\\Portmaster\\upgrade_obsolete_binaries: Incorrect function."`

PR #1869, opened 2025-04-14
Fix: updater restart action

stenya:
1. Fixed the actual action performed by the "Restart" button after an update
2. Updated configuration to preserve uninstaller.exe on Windows
3. Removed unnecessary artifacts from the updated "Ignore files" configuration

@dhaavi There’s a topic to discuss:
Currently, the UI simply closes when the "Restart" action is triggered.
We need to consider how to properly restart it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,3 +57,7 @@ windows_core_dll/.vs/windows_core_dll
 #windows_core_dll
 windows_core_dll/x64/
 windows_core_dll/portmaster-core/x64/
+
+#Binaries used for installer gereneration for Windows
+desktop/tauri/src-tauri/binary/
+desktop/tauri/src-tauri/intel/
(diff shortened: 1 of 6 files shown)


dhaavi:
In v2, Portmaster triggers the restart via the system service manager. Implementation is here: https://github.com/safing/portmaster/blob/v2.0/cmds/cmdbase/service.go

Rundown:
- Portmaster triggers shutdown/restart internally.
- Service manager (cmdbase) waits for shutdown to complete.
- If Portmaster service instance's `ShouldRestartIsSet()` returns true, the service manager tells the system service manager to restart the service.

stenya:
Thanks.
There are no problems with restarting the service.

I was referring to the UI app.
From the user's perspective, after pressing the "Restart" button, the UI simply closes and they have to start it manually.
I’m just considering the possibility of restarting the UI app automatically.

dhaavi:
Ah, I see. This should be able work just like in v1. The UI closes because it receives a shutdown event, it should be just receiving a restart event.

PR #1886, opened 2025-05-16
Fix/v2 force UI use ipv4 api

stenya:


dhaavi:
This will work for all usual installs.

However, it is possible to run Portmaster on a server and then expose the UI via a proxy or VPN to a client. This PR will make these cases fail, as the API must connect to the same host as the UI was loaded from - not to localhost.

This is not a common use case - though I know someone who uses/used Portmaster that way.

And wiki actually has a section on how to do this: https://wiki.safing.io/en/FAQ/RemotePortmaster

I think this should also work if you force `127.0.0.1:817` only within Tauri, and then only use ${window.location.host} everywhere in angular.

stenya:
@dhaavi The functionality you mentioned hasn't changed - remote connection to Portmaster is still possible.
The Angular production configuration remains unchanged: https://github.com/safing/portmaster/blob/v2.0/desktop/angular/src/environments/environment.prod.ts

However, you're right — it was sufficient to make the changes in the Tauri project only.
So I reverted the Angular changes, keeping just a minor logging improvement. https://github.com/safing/portmaster/commit/58ca3150e7b5d49f50d1f683b9b0383b26ef1725

PR #1893, opened 2025-05-22
Fix/restrict UI process connections

stenya:
Changes:

----

**1. The default configuration for the `Portmaster User Interface` profile was changed to block all connections.**
We need it to be sure that child WebView processes will not make any unexpected connections.

---- 
**2. Bugfix: `External processes wrongly classified as Portmaster UI on Windows`**
The `portmaster-core` uses specific logic to determine whether a process is related to the Portmaster UI binary.
If a process is a child of `C:\Program Files\Portmaster\portmaster.exe`, it is trusted as the Portmaster UI.
This allows `portmaster-core` to correctly recognize child WebView processes as part of the Portmaster UI.

Connections from the Portmaster UI are associated with a specific app profile: `Portmaster User Interface`.

**Problem:**
In some cases, the system browser may be launched from the Portmaster UI (e.g., when a user clicks a link in the UI).
Since the browser is started as a child of the Portmaster UI process, it is incorrectly trusted as part of the Portmaster UI.
As a result, all browser connections are treated as if they belong to the Portmaster User Interface profile, and the corresponding rules are applied.

**Steps to Reproduce (in PM version 2.0.14):**
1. Make sure no web browser processes are running.
2. Launch the Portmaster app.
3. Switch the UI to Developer Interface.
4. Click any link from the NEWS section (top-right corner of the app window).

**Observed Result:**
A new instance of the system's default web browser is started to open the link.
The Portmaster User Interface profile appears in the app list.
All web browser connections are listed under the Portmaster User Interface profile.
**Expected Result:**
The Portmaster User Interface profile should show only its own connections.
Web browser connections should be correctly associated with their own profiles (e.g., Google Chrome).

### DIFF
--- a/service/process/profile.go
+++ b/service/process/profile.go
@@ -118,6 +118,8 @@ func (p *Process) IsPortmasterUi(ctx context.Context) bool {
 	var previousPid int
 	proc := p
 
+	hasPmWebviewEnvVar := false
+
 	for i := 0; i < checkLevels; i++ {
 		if proc.Pid == UnidentifiedProcessID || proc.Pid == SystemProcessID {
 			break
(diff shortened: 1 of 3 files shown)


service/process/profile.go
```diff
@@ -118,14 +118,41 @@ func (p *Process) IsPortmasterUi(ctx context.Context) bool {
...
+			// If the env var is set, we are a child (WebView window) of the Portmaster UI process.
+			// Otherwise, the process was launched by the Portmaster UI, but should not be trusted as the Portmaster UI process.
+			if i == 0 {
+				return true // We are the main Portmaster UI process.
+			}
+			if hasPmWebviewEnvVar {
+				return true // We are a WebView window of the Portmaster UI process.
+			}
+			// The process was launched by the Portmaster UI, but should not be trusted as the Portmaster UI process.
+			log.Tracer(ctx).Warning(fmt.Sprintf("process: %d '%s' is a child of the Portmaster UI, but does not have the PORTMASTER_UI_WEBVIEW_PROCESS environment variable set. Ignoring.", p.Pid, p.Path))
```
dhaavi, line 149: You can directly use `Warningf(` instead of `.Warning(fmt.Sprintf(`.

Also, use `%q` instead of `'%s'`.

service/process/profile.go
```diff
@@ -118,14 +118,41 @@ func (p *Process) IsPortmasterUi(ctx context.Context) bool {
...
+			return false
+		}
+
+		// Check if the process has the environment variable set.
+		if _, ok := proc.Env["PORTMASTER_UI_WEBVIEW_PROCESS"]; ok {
+			hasPmWebviewEnvVar = true
 		}
```
dhaavi, line 156: You check the existence of the env var for the next iteration.
I am not sure how this will work out. If this is intentional, please document this, so this is clear for the future.

dhaavi (changes requested):
This is a good idea with the env var.

I only reviewed the code, not the implications on the detection of the UI processes.

service/process/profile.go
stenya: Changed

PR #1908, opened 2025-06-10
Kext bugfixes

vlabo:
1. 10f2ef426e2042a042de88531042e52175a20e35 Adds extra step to the shutdown sequence of the kext that will pop all connections that are waiting for a verdict and sets a verdict. That way the ALE layer will shutdown properly.
This fixes rare bug which happens if portmaster starts to shutdown and there are connections that dont have a verdict set yet, the kext will fail to unload and the process will hang.

2. 934480da0ba2886b5b080b07e2ee07af94e35864 Fixes the call to `copy_nonoverlapping` (aka `copy_from_slice`), on some scenarios was called on the same memory for source and destination, that is undefined behavior.


dhaavi (dismissed):
LGTM

stenya (approved):

stenya (approved):

PR #2122, opened 2026-01-26
Packet injection and verdict setting fixes.

vlabo:
61c3443deb4a83ca3dab2dac7fd1533456ab35b1: Makes the block and drop verdict immutable. This will make windows apply the verdict immediately and not let other layers in the chain override the verdict. Also makes the applying of the verdict immediate in the ALE layer. So applications will get permission denied when trying to create a connection instead of timeout after few seconds. 

04f8708448439fbef3f70a89b7d7714fdfb4375b: Fixes the packet injection in the packet callout. From what I remember this is only used in packet dns monitoring. Without recalculating the checksums the packet will just get dropped after injection by the network system.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed incorrect checksum calculations for network packets in outbound traffic, ensuring proper packet validation for both IPv4 and IPv6 protocols.

### DIFF
--- a/windows_kext/driver/src/packet_callouts.rs
+++ b/windows_kext/driver/src/packet_callouts.rs
@@ -274,6 +274,12 @@ fn clone_packet(
     };
 
     if let Some(data) = clone.get_data_mut() {
+        // Outbound packets intercepted at the IP layer may carry only a partial
+        // pseudo-header checksum because the TCP/IP stack relies on NIC hardware
+        // checksum offload to fill in the real value before transmission.
+        // When this clone is later re-injected via FwpsInjectNetwork*Async (on
+        // Accept/PermanentAccept verdict), it bypasses the NIC entirely, so offload
+        // never runs. We must compute the full software checksum here.
         recalc_header_checksums(data, ipv6);
     }
 
--- a/windows_kext/driver/src/packet_callouts.rs
+++ b/windows_kext/driver/src/packet_callouts.rs
@@ -12,7 +12,9 @@ use crate::connection::{
 use crate::connection_cache::ConnectionCache;
 use crate::connection_map::Key;
 use crate::device::{Device, Packet};
-use crate::packet_util::{get_key_from_nbl_v4, get_key_from_nbl_v6, Redirect};
+use crate::packet_util::{
+    get_key_from_nbl_v4, get_key_from_nbl_v6, recalc_header_checksums, Redirect,
+};
 
 // IP packet layers
 pub fn ip_packet_layer_outbound_v4(data: CalloutData) {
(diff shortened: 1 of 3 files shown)


windows_kext/wdk/src/filter_engine/callout_data.rs
```diff
@@ -188,6 +190,8 @@ impl<'a> CalloutData<'a> {
 
     pub fn block_and_absorb(&mut self) {
         unsafe {
+            // Make the verdict immutable so the next layer can't change it
```
stenya: It seems that `clear_write_flag()` is redundant when `set_absorb()` is called.

windows_kext/wdk/src/filter_engine/callout_data.rs
```diff
@@ -174,6 +174,8 @@ impl<'a> CalloutData<'a> {
 
     pub fn action_block(&mut self) {
         unsafe {
+            // Make the verdict immutable so the next layer can't change it
```
stenya: This commit will conflict with commit https://github.com/safing/portmaster/pull/2136/changes/35ecee89d5736a726f74c334f41ee1aa333a246f, which is doing the same thing.
Could you please revert it so that PR https://github.com/safing/portmaster/pull/2136 does not conflict with yours?

stenya:
Hello @vlabo 

I’m sorry for the long delay in reviewing this - I only just noticed your pull request.
Many thanks for your efforts.

I only have a few comments:
- https://github.com/safing/portmaster/commit/61c3443deb4a83ca3dab2dac7fd1533456ab35b1  :
This commit will conflict with commit https://github.com/safing/portmaster/pull/2136/changes/35ecee89d5736a726f74c334f41ee1aa333a246f, which is doing the same thing.
Could you please revert it so that PR https://github.com/safing/portmaster/pull/2136 does not conflict with yours?

- Regarding https://github.com/safing/portmaster/commit/04f8708448439fbef3f70a89b7d7714fdfb4375b:
Could you please add a comment explaining the change and why it is needed?
Something like this (if I understood everything correctly):
```
// Checksums may be zero/garbage due to NIC hardware offload. When injecting
// via software those flags are absent, so we must write valid checksum bytes
// before storing the clone — even if the packet is otherwise unmodified.
```
---
By the way, I would really appreciate any feedback you might have on these commits: https://github.com/safing/portmaster/pull/2136/changes/674ff3f4dc9fcedb73ea1b1a12e9629de581be74, https://github.com/safing/portmaster/pull/2136/changes/2d6fd0ad09b08c0fc6352d998153c9b06573e0da,
or on the whole PR https://github.com/safing/portmaster/pull/2136 in general.

windows_kext/driver/src/packet_callouts.rs
```diff
@@ -265,11 +267,16 @@ fn clone_packet(
     interface_index: u32,
     sub_interface_index: u32,
 ) -> Result<Packet, String> {
-    let clone = nbl.clone(&device.network_allocator)?;
+    let mut clone = nbl.clone(&device.network_allocator)?;
     let inbound = match direction {
         Direction::Outbound => false,
         Direction::Inbound => true,
     };
+
```
stenya, line 275: Could you please add a comment explaining the change and why it is needed?
Something like this (if I understood everything correctly):
```
// Checksums may be zero/garbage due to NIC hardware offload. When injecting
// via software those flags are absent, so we must write valid checksum bytes
// before storing the clone — even if the packet is otherwise unmodified.
```

stenya (approved):
Thank you!

vlabo:
Sorry for the long delay. For some reason I did not get a notification.
I just reviewed the changes and they look good.

stenya:
Thank you @vlabo !

PR #2217, opened 2026-08-07
fix(windows): preserve unrelated startup entries on uninstall

### LINKED ISSUE #2140: Bug: Portmaster Uninstaller Permanently Breaks Windows Security Taskbar Icon, removes other startup entries
Using latest release: v2.1.7, Windows 25H2.

Using portmasters uninstaller destroys startup entries not related to portmaster and permanently damages windows functions.  SFC and DISM do not detect or fix it. Only a restore point or image can fix this problem, possibly a in place dist-upgarde. 

Issues:

Windows security taskbar icon disappears: 
Image
This icon disappears and there is no way to bring it back, tried everything in window taskbar settings and nothing brings it back.

Other app startups fail:

Softperfect's Networx
Binisoft WFC (windows firewall control) UI does not load at startup.

The first of the latter two can be fixed by applying startup option in the settings AGAIN, and the second one by manually editing the registry.

**How did you reproduce it?**:

I restored windows to a previous state when portmaster was installed... and found there was no issue until I uninstalled portmaster again. I repreoduced it a dozen times. I did another restore and tested Portmasters different uninstall methods.

1. Keeping user data and
2. Uninstalling AND removing user data. 

Both resulted with the same problem. The only way to remove portmaster without permanately damaging windows is manually uninstalling or in my case revo-uninstaller, which worked perfectly.

GCDBusiness:
## Summary

- remove only the Portmaster startup value during NSIS uninstall
- preserve unrelated machine-wide startup entries

## Why

The installer creates a named Portmaster value in the shared HKLM CurrentVersion Run key. The uninstaller currently deletes the entire shared key, which removes startup entries belonging to Windows and other applications. This changes the cleanup operation to delete only the value created by Portmaster.

## Testing

- git diff --check
- verified that the install and uninstall hooks use the same registry path and value name
- full NSIS bundle not built locally because makensis is unavailable

Fixes #2140

## Summary by CodeRabbit

* **Bug Fixes**
  * Uninstalling the application now removes only its autostart registry entry, preserving other applications’ startup settings.

### DIFF
--- a/desktop/tauri/src-tauri/templates/nsis/install_hooks.nsh
+++ b/desktop/tauri/src-tauri/templates/nsis/install_hooks.nsh
@@ -179,7 +179,7 @@ var dataDir
 
   ; remove the registry entry for the autostart
   DetailPrint "Removing registry entry for autostart"
-  DeleteRegKey HKLM "SOFTWARE\Microsoft\Windows\CurrentVersion\Run"
+  DeleteRegValue HKLM "SOFTWARE\Microsoft\Windows\CurrentVersion\Run" "Portmaster"
 
   ; delete data files
   Delete  /REBOOTOK "$COMMONPROGRAMDATA\Portmaster\databases\history.db"


stenya (approved):

stenya:
@GCDBusiness  Great finding!
Thank you!

PR #2267, opened 2026-09-13
Reduce the amount of binding errors during DNS queries

dannyagostinelli:
Portmaster picks its own pre-authenticated local port from 10000-65535, but the OS may have reserved parts of that range. Windows does this for Hyper-V, WSL and Docker. When `GetUnusedLocalPort` picks a reserved port the bind fails, the query never leaves the machine, and it still gets counted as a failure against the DNS server. This is the cause of the bind errors in #2066, diagnosed there by @JeffSchofield.

In this PR, `GetUnusedLocalPort` now confirms a candidate can actually be bound before returning it, and remembers the ones that couldn't for an hour so they aren't tried again right away. The entries expire because a bind can also fail for a port that something else held for a moment, and those should be retried. 

The check binds a UDP socket and closes it without sending anything, so it creates no traffic. This mechanism is purely local so there's no handshake, no packet, and it takes only about 60 microseconds from testing on my PC. The only way to truly validate a TCP source port is to connect with the port, which is a network round trip, so for TCP and DoT the dial is the check and the retry handles a failed bind. DoH doesn't use pre-authenticated ports, so I didn't implement anything for that. 

The resolvers also retry with a different port, up to three attempts, when the local bind fails. This covers any gap between checking a port and using it, TCP, and plain collisions where something takes the port first.

I looked at reading the reserved ranges from the OS instead, but Windows has no documented way to enumerate them and I thought other OSes would all handle this differently, making the solution hard to implement without testing many OS versions for their behavior. Also, any ranges can also change, so a cached copy would go stale anyway and so you'd still need a more robust bind and retry mechanism. Checking the port directly avoids all of that and is not Windows specific.

I did testing based on the proposal and included them. I measured with `portIsBindable` itself on Windows 11. I had 1860 ports reserved by WSL and Docker and tested over 5000 selections. About 3% of ports handed out could not be bound before the change and 0% after. The validation took about 63 microseconds per selection. The whole DNS query took between 10 to 500 milliseconds.

Tests cover the bind check, that TCP is left alone, and that remembered ports expire. The retry check matches on the failing syscall being `bind` rather than an errno, so it stays portable and only catches failures where nothing was sent. On Windows an excluded UDP port gives `&net.OpError{Op:"dial", Err:&os.SyscallError{Syscall:"bind", Err:0x271d}}` and an excluded TCP port the same, while a free port and a `connectex` refusal don't match.

Verifications completed:
- go vet ./service/network/ ./service/resolver/ : clean
- gofmt -l : clean
- go test ./service/network/ : ok (4 new tests pass)
- go test ./service/resolver/ -run TestIsLocalBindError : pass
- go build ./... and go test ./service/... : the set of failing packages is
  identical with and without the patch (ebpf packages are Linux only and do not
  build on Windows; geoip, resolver, updates, splittun/proxy and profile/endpoints
  need network access or a live service).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved DNS resolution reliability when local port binding fails by retrying with alternative ports.
  * Avoided selecting recently unusable UDP ports.
  * Added validation to ensure selected UDP ports can be bound before use.
  * Preserved existing behavior for TCP ports while improving handling of local bind failures.
  * Reduced failures caused by ports that are temporarily unavailable during DNS queries.


service/network/ports.go
Terrance, line 32: Rather than generating a random port and checking its availability, can we not just ask the system for a free port?  Binding port zero usually lets the system decide, and it looks like this is already done as a fallback:

https://github.com/safing/portmaster/blob/c6810cb66576b7ac930ede5a2504d09c580f2d59/service/firewall/preauth.go#L72-L74

dannyagostinelli: I think you're right and I like that better. We can bind to port 0, see which port the OS gave us, allow that port through the firewall, then send on that same socket. The OS shouldn't give out a reserved or busy port, so the port check, the cache, and the one hour timer can all go away. TCP can't do this because connect sends the first packet right away, so TCP would keep the retry.

@stenya Do you want me to rework this PR to do it that way instead? I've moved it to a draft while we sort it out.

stenya: hi. I’m hoping to get to this PR sometime next week.

dannyagostinelli:
No problem. The main question is do you want me to rework this to Terrance’s recommendation before you review it or keep it as-is? If you want me to rework it, let me know soon so I can have it ready for your review. On Sep 15, 2026, at 2:45 AM, Alexandr Stelnykovych ***@***.***> wrote:﻿
@stenya commented on this pull request.



In service/network/ports.go:
> 

hi. I’m hoping to get to this PR sometime next week.

service/network/ports.go
dannyagostinelli: No problem. The main question is do you want me to rework this to Terrance’s recommendation before you review it or keep it as-is? If you want me to rework it, let me know soon so I can have it ready for your review.

